PR #3, opened 2012-12-02
Feature/xwiki rendering html5

ElderMael:
HTML5 Parser and Renderer, please refer to initial SO question:

http://stackoverflow.com/questions/13569645/creole-1-0-to-html5-renderer-for-java

Any comments or reviews appreciated.


gdelhumeau:
Hi @ElderMael.

Thanks your for you pull request. It's a pity we haven't found a way to improve your contribution together, but our last message went unanswered for so long now and the HTML5 renderer has been added in the meantime by an other contributor (me :)).

I think there is no need to let this pull request open now. If you wish to improve the implementation we have in XWiki, you are warmly welcome.

Thanks,
Guillaume

PR #6, opened 2013-06-27
Preventing XSS in wiki syntax

tdelafosse:
I introduced new "secure" renderers and printers that can be used in xwiki-platform to prevent XSS. See also the corresponding pull request made on xwiki-platform, that makes use of this.


michitux:
This has been implemented differently as part of XRENDERING-663/#212.

PR #12, opened 2016-03-17
Fix the typo in the name of ListBlock

woshilapin:
This change is [Reviewable](https://reviewable.io/reviews/xwiki/xwiki-rendering/12)

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ListBlock.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ListBlock.java
@@ -23,11 +23,9 @@ package org.xwiki.rendering.block;
  * Represents any type of Lists (numbered list, bulleted list, etc).
  *
  * @version $Id$
- * @since 1.5M2
- * @deprecated since 8.1M1, use {@link ListBlock} instead
+ * @since 8.1M1
  */
-@Deprecated
-public interface ListBLock extends ListBlock
+public interface ListBlock extends Block
 {
     // This is currently only a tagging interface and thus it has no methods.
 }
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
@@ -33,7 +33,7 @@ import org.xwiki.rendering.block.Block;
 import org.xwiki.rendering.block.BulletedListBlock;
 import org.xwiki.rendering.block.HeaderBlock;
 import org.xwiki.rendering.block.LinkBlock;
-import org.xwiki.rendering.block.ListBLock;
+import org.xwiki.rendering.block.ListBlock;
 import org.xwiki.rendering.block.ListItemBlock;
 import org.xwiki.rendering.block.NumberedListBlock;
 import org.xwiki.rendering.block.SectionBlock;
@@ -200,7 +200,7 @@ public class TocMacro extends AbstractMacro<TocMacroParameters>
 
                 if (currentLevel < headerLevel) {
                     while (currentLevel < headerLevel) {
-                        if (currentBlock instanceof ListBLock) {
+                        if (currentBlock instanceof ListBlock) {
                             currentBlock = addItemBlock(currentBlock, null);
                         }
 
@@ -274,9 +274,9 @@ public class TocMacro extends AbstractMacro<TocMacroParameters>
      * @param parentBlock the block where to add the new list block.
      * @return the new list block.
      */
-    private ListBLock createChildListBlock(boolean numbered, Block parentBlock)
+    private ListBlock createChildListBlock(boolean numbered, Block parentBlock)
     {
-        ListBLock childListBlock =
+        ListBlock childListBlock =
             numbered ? new NumberedListBlock(Collections.<Block>emptyList()) : new BulletedListBlock(
                 Collections.<Block>emptyList());
 


vmassol:
Thanks Jean but AFAICS this should have broken clirr when you built it with maven :)

tmortagne:
Yes this is an important API breakage, any code searching for lists in a XDOM will be broken.

woshilapin:
I had problem with other tests on Jira Macro so I disconnect the tests.  I made this PR because I lost a lot of time in Groovy (no autocompletion or compilation) finding an error which was only this typo.  But indeed, it's an API change.

woshilapin:
Does this all means that the best is that this error lies here forever?

vmassol:
No it means you need to find a solution that doesn't break backward compat if you want to provide a PR :)

The solution is to deprecate ListBLock and introduce a ListBlock one and handle both.

vmassol:
It may not be that easy actually. Imagine I have existing code looking for all ListBLock blocks. Even if we handle both in the traversal methods, we won't be be able to return ListBLock or ListBlock since it would be typed...

vmassol:
This needs to be analyzed, it's probably quite complex. We would need to do an impact analysis and see if there's something we can break with minimal impact on users and on devs.

tmortagne:
You can deprecate ListBLock, make it implement ListBlock and keep extending ListBLock in other blocks (otherwise someone searching for ListBLock will never find anything).

woshilapin:
I'll try that and keep you post.

woshilapin:
OK, I've made the modifications. First time I do deprecation (or try too) so I hope it's fine.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ListBLock.java
```diff
@@ -24,7 +24,9 @@
  *
  * @version $Id$
  * @since 1.5M2
+ * @deprecated Use {@link ListBlock} instead
```
tmortagne: @deprecated since 8.1M1, use {@link ListBlock} instead

since 8.1M1 will be the next release in this branch

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ListBlock.java
```diff
@@ -0,0 +1,31 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.rendering.block;
+
+/**
+ * Represents any type of Lists (numbered list, bulleted list, etc).
+ *
+ * @version $Id$
+ * @since 8.1
```
tmortagne: 8.1M1

woshilapin:
OK, I've made the modifications on JavaDoc.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ListBLock.java
```diff
@@ -24,7 +24,9 @@
  *
  * @version $Id$
  * @since 1.5M2
+ * @deprecated since 8.1M1, use {@link ListBlock} instead
  */
+@Deprecated
 public interface ListBLock extends Block
```
tmortagne: As I indicated you need to extend the new ListBlock here. It automatically fix all implementations of ListBLock, even those you don't know.

woshilapin:
Ah right, I misunderstood and indeed, it's wiser. Just did it!

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/DefinitionListBlock.java
```diff
@@ -30,7 +30,7 @@
  * @version $Id$
  * @since 1.6M2
  */
-public class DefinitionListBlock extends AbstractBlock implements ListBLock
+public class DefinitionListBlock extends AbstractBlock implements ListBlock
```
tmortagne: I guess you removed the wrong one here.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/BulletedListBlock.java
```diff
@@ -31,7 +31,7 @@
  * @version $Id$
  * @since 1.5M2
  */
-public class BulletedListBlock extends AbstractBlock implements ListBLock
+public class BulletedListBlock extends AbstractBlock implements ListBlock
```
tmortagne: I guess you removed the wrong one here.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ListBLock.java
```diff
@@ -24,8 +24,10 @@
  *
  * @version $Id$
  * @since 1.5M2
+ * @deprecated since 8.1M1, use {@link ListBlock} instead
  */
-public interface ListBLock extends Block
+@Deprecated
+public interface ListBLock extends Block, ListBlock
```
tmortagne: Just a detail: ListBlock is a Block already :)

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/NumberedListBlock.java
```diff
@@ -31,7 +31,7 @@
  * @version $Id$
  * @since 1.5M2
  */
-public class NumberedListBlock extends AbstractBlock implements ListBLock
+public class NumberedListBlock extends AbstractBlock implements ListBlock
```
tmortagne: I guess you removed the wrong one here. Should be ListBLock (otherwise old code searching for ListBLock won't find it).

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ListBLock.java
woshilapin: Yes, sure...

woshilapin:
Shouldn't have do that in a hurry. I realize that I didn't fully understood what you were trying to explain to me because it means that all implementations will still extends the deprecated class. I thought that we would remove any mention of `ListBLock` from our code, except keeping this deprecated class for any other external code that would still depend on it.

Could we not instead make `ListBlock` extends `ListBLock` and remove any mention of `ListBLock` everywhere else. This way, the day we remove completely this class, it's just about removing the `extends` from `ListBlock` and removing the class `ListBLock`? Or would it mean again modifying an API and that why we don't do like that?

In the meantime, I've applied the modification that (I think :-P) I've understand now so you can merge if you prefer this way.

tmortagne:
Retro compatibility is usually more about implementations you don't control that implementation you control. The point of making the deprecated interface extends the new interface is to make all implementations (even those you don't know) implement the new interface and be found when you search for the new interface.

woshilapin:
Would you like a Jira for this one and a modification of the commit message?

tmortagne:
Indeed I did not noticed but would be clear to associate this change to a jira issue, after all there is a new API in it.

woshilapin:
Just did it.

tmortagne:
Applied. Thanks for this not so easy contribution !

PR #14, opened 2017-12-01
Update AbstractBlock.java

cweckle:
This is my submission for the "Fix a coding violation" task.


vmassol:
Nice, thanks! 

In the future please try to write a commit comment that explains why something is done instead of what is being done. For example instead of "Update AbstractBlock.java" which is not very meaningful you could have said "Improve readability/maintainability of code" (which is the reason for preferring `isEmpty()` to `size() == 0`.

PR #15, opened 2017-12-01
[java] Logic related reliability

MargaretM22:
Objects should be compared with "equals()"


vmassol:
Hi, thanks. However this change is a breaking change which changes the logic. The code is explictly comparing the object's references. As you can see the test is failing for this reason.

So this means that this is a false positive raised by sonarqube. We should mark it as such in sonarqube.

PR #16, opened 2017-12-02
[Java] Logic Related Reliability

MargaretM22:
Call Super in the constructor


aubincleme:
Thanks for your contribution!
For the next time, please start your commits with the prefix `[misc]` in order to signify that you are not committing something related to a JIRA issue.
Here is an example of a commit having the `[misc]` prefix : https://github.com/xwiki/xwiki-platform/commit/c47186dcfd839cf5e18119b898e1a3178c5dbf3a

PR #17, opened 2017-12-06
[codestyle] use equals method to compare 

akshitdewan:


tmortagne:
As explained in the method javadoc we explicitly want to compare reference et not equality here.

PR #18, opened 2017-12-06
[codestyle] replaced printStackTrace() with logger

akshitdewan:


xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/tex/TexSerializer.java
```diff
@@ -465,7 +467,7 @@ protected InputStream getImageInput(String ref) throws IOException
                 WGet wget = new WGet();
                 wget.fetchURL(url, fos);
             } catch (Exception e) {
-                e.printStackTrace();
+                LOGGER.error("Error while getting image input",e);;
```
vmassol: This class is an old class used by wikimodel that we integrated and don't use. AFAICS it's not finished work and uses System.out/err in several places. I think we might be better off simply removing this class. Just need to check that we really don't use it (we have a latex Renderer AFAIR).

Small note: you forgot a space after the commas.

Also, when writing debug logs you need to provide the max relevant info so that it can be debugged easily. For example you forgot to add the URL or some kind of image reference to identify the image input.

Thanks!

xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/tex/TexSerializer.java
```diff
@@ -42,13 +42,15 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 /**
  * @version $Id$
  * @since 4.0M1
  */
 public class TexSerializer extends PrintTextListener
 {
+    private static final Logger LOGGER = LoggerFactory.getLogger(TexSerializer.class);
     private static class DocumentContext
```
tmortagne: You are missing an empty line after Logger declaration. Would be a pity to not follow XWiki codestyle for a [codestyle] improvement :)

If you use the Eclipse in the dev VM you can use Ctrl+Shift+F to properly format code (then make sure to commit only what's related to your work, sometime the file is not full following XWiki formatting rules in other places).

xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/tex/TexSerializer.java
```diff
@@ -167,9 +169,9 @@ public final void readURL(URLConnection url) throws IOException
                     System.out.println((char) in.readUnsignedByte());
                 }
             } catch (EOFException e) {
-                e.printStackTrace();
+                LOGGER.error("EOF Exception while reading url : ",e);
```
tmortagne: Don't put `:`, just detail what went wrong at this level (and include the URL you failed to read). The cause exception will be printed after.

vmassol: BTW the best practices on logging is available at http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle#HLoggingBestPractices

xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/tex/TexSerializer.java
```diff
@@ -161,15 +165,15 @@ public final void readURL(URLConnection url) throws IOException
         {
             DataInputStream in = new DataInputStream(url.getInputStream());
             printHeader(url);
-
+
```
tmortagne: This looks like a mistake.

tmortagne:
Just a minor leftover to fix and you are good to go :)

tmortagne:

PR #19, opened 2017-12-07
[codestyle] works better with J2EE

akshitdewan:
Might be surprised by how classLoader() works with J2EE, improves some test cases

### DIFF
--- a/xwiki-rendering-test/src/main/java/org/xwiki/rendering/test/cts/TestDataParser.java
+++ b/xwiki-rendering-test/src/main/java/org/xwiki/rendering/test/cts/TestDataParser.java
@@ -75,7 +75,7 @@ public class TestDataParser
     public List<TestData> parseTestData(String syntaxId, String ctsRootPackageName, String packageFilter,
         String pattern) throws Exception
     {
-        ClassLoader classLoader = getClass().getClassLoader();
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
 
         List<TestData> data = new ArrayList<TestData>();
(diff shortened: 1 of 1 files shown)


xwiki-rendering-test/src/main/java/org/xwiki/rendering/test/cts/TestDataParser.java
```diff
@@ -75,7 +75,7 @@
     public List<TestData> parseTestData(String syntaxId, String ctsRootPackageName, String packageFilter,
         String pattern) throws Exception
     {
-        ClassLoader classLoader = getClass().getClassLoader();
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
```
vmassol: hmm I don't think this is right. The idea was to use the classloader used by the TestDataParser class to load the other classes. This is used when tests are executing. I don't think the current context CL is even set at that time. Did you try the change in real?

vmassol: (this is not JavaEE code :))

tmortagne: This time this change is a mistake. We explicitly want to make sure to use the classloader of the test class to make sure to not end up with unexpected class.

In other words you found a false positive.

akshitdewan: alright I'll put it on the list

PR #20, opened 2017-12-07
[codestyle] Renamed isInline to inline in MacroMarkerBlock

TheReturningVoid:
Fixes http://sonar.xwiki.org/issues/search#issues=0b90d4fd-bb45-4860-a4d3-a422e0fc14c3


tmortagne:

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/MacroMarkerBlock.java
```diff
@@ -49,7 +49,7 @@
     /**
      * The macro is located in a inline content (like paragraph, etc.).
      */
-    private boolean isInline;
+    private boolean inline;
```
vmassol: Thx. Personally I was finding isInline a better name and more clear and following our naming conventions that we  use. So I don't consider this an improvement.

tmortagne: How is this our naming convention ? "isInline" is supposed to be the method name, not the field.

You are not going to find many boolean fields named like this.

vmassol: So you disagree that "isInline" is a better and more clear name than "inline"?

Re naming conventions we don't have anything written (for method or field names) but I've personally followed this convention from the beginning whenever I could, using "isXXX" or "shouldXXX". We have that in lots of places. Would be too long to list here. Just showing one recent example:

```java
public class DefaultNotificationFilterPreference implements NotificationFilterPreference
{
    private boolean isEnabled;

    private boolean isActive;
```

tmortagne: * Having the method be named "isInline" and the field "inline" is quite a Java standard (as you can see from Sonar complains) and as you said there is nothing in our convention that validate your "our naming conventions" statement
* From what I can see there is about 6 times more boolean following the Java standard in commons/rendering/platform so even in practice we could hardly call that "our naming convention"

mflorea: FTR, I've always used "enabled" as field name and not "isEnabled". For me the "is" prefix is redundant. It's quite clear from the name "enabled" that it's a boolean field. As for the method name, there's no doubt that "isEnabled()" is the standard.

vmassol: ok interesting. I fail to understand why it would be obvious for a field name but not for a method name but so be it :)

vmassol: I still don't agree but I'm agreeing to disagree :)

vmassol: which means that FTM we'll have two conventions for naming variables.

PR #21, opened 2017-12-07
[codestyle] added hashcode method

akshitdewan:
hashcode method supplements the equals method

### DIFF
--- a/xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/util/ListBuilder.java
+++ b/xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/util/ListBuilder.java
@@ -64,7 +64,7 @@ public class ListBuilder
         @Override 
         public int hashCode()
         {
-            return new HashCodeBuilder(3,17)
+            return new HashCodeBuilder()
                 .append(fTreeChar)
                 .append(fPos)
                 .toHashCode();
--- a/xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/util/ListBuilder.java
+++ b/xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/util/ListBuilder.java
@@ -22,6 +22,8 @@ package org.xwiki.rendering.wikimodel.util;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
 /**
  * This is an internal utility class used as a context to keep in memory the
  * current state of parsed trees (list items).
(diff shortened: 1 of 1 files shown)


xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/util/ListBuilder.java
```diff
@@ -58,6 +60,16 @@ public boolean equals(Object obj)
             CharPos pos = (CharPos) obj;
             return equalsData(pos) && pos.fPos == fPos;
         }
+        
+        @Override 
+        public int hashCode()
+        {
+            return new HashCodeBuilder(3,17)
```
tmortagne: Why don't you simply use the empty constructor ?

vmassol: Because that's what we document as our best practice: http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle#HEquals2FHashCodeandToStringimplementations

What happens when you use the empty one? How are the magic numbers computed? (would need to read the javadoc). If it's better then we need to update our doc and code. I also use magic numbers (that I try to change for each class to increase "randomness").

vmassol: AFAIK this is also what's documented at http://commons.apache.org/proper/commons-lang//apidocs/org/apache/commons/lang3/builder/HashCodeBuilder.html

vmassol: "Uses two hard coded choices for the constants needed to build a hashCode." not the best.

vmassol: maybe acceptable? Would need to understand better to know if it's acceptable. Let me know if you have.

tmortagne: I don't really see the point exactly most of the time, you are not really supposed to compare the hash code of two different kind of objects.

tmortagne: > AFAIK this is also what's documented at http://commons.apache.org/proper/commons-lang//apidocs/org/apache/commons/lang3/builder/HashCodeBuilder.html

The constructor javadoc also say "Ideally these should be different for each class, however this is not vital." :) Sure this is probably a bit safer but what is sure is that always setting the same thing is totally useless (even if you try it's not so easy to maintain different numbers across all classes with various developers)  :)

tmortagne: Proposal sent :)

tmortagne:

PR #22, opened 2017-12-07
[codestyle] removed unnecessary constructor

akshitdewan:


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/NewLineBlock.java
```diff
@@ -37,10 +37,6 @@
      *
      * @since 2.6RC1
      */
-    public NewLineBlock()
-    {
-        // We need to keep this constructor to override the ones in AbstractBlock
```
vmassol: Could you explain why you're removing it when a comment says to not remove it? :) Is the comment wrong? If so, why? Thx

tmortagne: Yes the comment is wrong:
* this constructor does not prevent the super class one to be called at all, that's not how object programming works
* the super class one does not do anything anyway so why would be need to "override" it ?

tmortagne:

PR #23, opened 2017-12-07
[codestyle] changed locale to english

akshitdewan:
toUpperCase() and toLowerCase() can cause problems with other languages such as Turkish. This is because toLowerCase() actually calls toLowerCase(Locale.default) and the default locale varies based on location.

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/listener/descriptor/DefaultListenerDescriptorManager.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/listener/descriptor/DefaultListenerDescriptorManager.java
@@ -21,6 +21,7 @@ package org.xwiki.rendering.internal.listener.descriptor;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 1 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/listener/descriptor/DefaultListenerDescriptorManager.java
```diff
@@ -115,7 +116,7 @@ public ListenerDescriptor createDescriptor(Class<?> type)
      */
     private void addElement(String elementName, ListenerDescriptor descriptor, Method method)
     {
-        String lowerElementName = elementName.toLowerCase();
+        String lowerElementName = elementName.toLowerCase(Locale.ENGLISH);
```
vmassol: Shouldn't we use Locale.ROOT instead?

See https://garygregory.wordpress.com/2015/11/03/java-lowercase-conversion-turkey/

tmortagne: That makes Locale.ROOT a lot less neutral than I tough if it's dealing with latin letters uppercase/lowercase :) I find it a bit counter intuitive.

tmortagne: Anyway if we are sure this is officially supported and Local.ROOT won't suddenly stop doing that the I gree it make more sense for use cases like these (since it's not exactly english content but method names).

tmortagne:

PR #25, opened 2017-12-08
[codestyle] changed size==0 to is empty

akshitdewan:


tmortagne:

PR #26, opened 2017-12-08
[codestyle] removed unnecessary constructor

akshitdewan:


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/SpaceBlock.java
```diff
@@ -36,10 +36,6 @@
      *
      * @since 2.6RC1
      */
```
tmortagne: You need to remove the javadoc of the constructor too.

akshitdewan:
Thomas, I removed the part you highlighted, not sure if you want to remove whats above that because it seems like an important description of a SpaceBlock.

tmortagne:
This was is the constructor javadoc so it does not make any sense to keep it there. You should move the part related to SpaceBlock in the class javadoc.

akshitdewan:
Moved them

tmortagne:

PR #27, opened 2017-12-09
[codestyle] replaced unnecessary loop with clear method

akshitdewan:


akshitdewan:
fixed

xwiki-rendering-macros/xwiki-rendering-macro-footnotes/src/main/java/org/xwiki/rendering/internal/macro/footnote/PutFootnotesMacro.java
```diff
@@ -175,9 +175,8 @@ private ListItemBlock processFootnote(MacroMarkerBlock footnoteMacro, int counte
     private void addFootnoteRef(MacroMarkerBlock footnoteMacro, Block footnoteRef)
     {
         for (ListIterator<Block> it = footnoteMacro.getChildren().listIterator(); it.hasNext();) {
```
tmortagne: Seems to be this whole code is equivalent to `footnoteMacro.getChildren().clear();`

akshitdewan:
Fixed

tmortagne:
Still this whole method does not make much sense. It's documented as adding something when in fact it's replacing everything.

Let's say the good side if that your change makes it very clear how weird this method is :)

tmortagne:

PR #31, opened 2017-12-09
[codestyle] isEmpty instead of size

akshitdewan:
http://sonar.xwiki.org/issues/search#issues=21f1f5be-ef7a-4604-aa4e-10dd89071433
http://sonar.xwiki.org/issues/search#issues=67c68008-c948-4430-8d48-adde88d29582


tmortagne:
Thanks for your contribution

PR #33, opened 2017-12-10
[codestyle] Change the comparison to use the equals method

SrijanJha28:
Update XHTMLChainingRenderer.java by Changing the comparison to use the equals method.


tmortagne:

PR #34, opened 2017-12-10
 [codestyle] fix inefficient char to String conversion

CandyZQ:
[codestyle] fix inefficient char to String conversion


tmortagne:

PR #35, opened 2017-12-11
[codestyle] Remove duplicate block of code

JL709:
Removed hashCode() due it being a duplicate method already existing in MacroMarkerBlock.


vmassol:
Hi there. This removal looks invalid since it implements hashcode for the current class (which has a different identity than its parent).

tmortagne:
Yes as @vmassol said this is not the right fix for this issue. This code is not useless, it's just very similar to the one in MacroMarkerBlock and they should be shared (for example trough an AbstractMacroBlock).

tmortagne:
Closing since the corresponding GCI task has been abandoned. Don't hesitate to reopen if you want to work on it.

PR #36, opened 2017-12-11
[codestyle] Reduce the number of returns

piyush69:
Reduce the number of returns of this method 4, down to the maximum allowed 3


xwiki-rendering-macros/xwiki-rendering-macro-ctsreport/src/main/java/org/xwiki/rendering/internal/macro/ctsreport/Result.java
```diff
@@ -59,15 +59,12 @@ public String toString()
     @Override
     public boolean equals(Object object)
     {
-        if (object == null) {
+        if (object == null || object.getClass() != getClass()) {
```
tmortagne: Since Result is not really designed to be extended, a slight improvement would be

    if (!(object instanceof Result))

(it cannot be an instance of Result if it's null)

tmortagne:

PR #37, opened 2017-12-11
 [codestyle] remove unused variable and redundant initializers

CandyZQ:


xwiki-rendering-macros/xwiki-rendering-macro-footnotes/src/main/java/org/xwiki/rendering/internal/macro/footnote/PutFootnotesMacro.java
```diff
@@ -175,9 +175,7 @@ private ListItemBlock processFootnote(MacroMarkerBlock footnoteMacro, int counte
     private void addFootnoteRef(MacroMarkerBlock footnoteMacro, Block footnoteRef)
     {
         for (ListIterator<Block> it = footnoteMacro.getChildren().listIterator(); it.hasNext();) {
-            Block b = it.next();
```
mflorea: Are you sure you can call ``remove()`` without calling ``next()`` before? Is the iterator moving to the next element when you call ``remove()``? Check the JavaDoc of [``Iterator.remove()``](https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#remove--).

tmortagne:
Akshit Dewan is already working on this see https://github.com/xwiki/xwiki-rendering/pull/27.

PR #38, opened 2017-12-11

SrijanJha28:
Update Test.java by Reducing the number of returns of this method 4, down to 3.


xwiki-rendering-macros/xwiki-rendering-macro-ctsreport/src/main/java/org/xwiki/rendering/internal/macro/ctsreport/Test.java
```diff
@@ -65,15 +65,12 @@ public String toString()
     @Override
     public boolean equals(Object object)
     {
-        if (object == null) {
+        if (object == null || (object.getClass() != getClass())) {
```
tmortagne: Since Test is not really designed to be extended, a slight improvement would be

    if (!(object instanceof Test))

(it cannot be an instance of Test if it's null)

PR #40, opened 2017-12-11

SrijanJha28:
Update Test.java by Reducing the number of returns of this method 4, down to 3.


tmortagne:

PR #42, opened 2017-12-11

SrijanJha28:
Update XWikiSyntaxChainingRenderer.java by Adding a default case to the switch.


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/main/java/org/xwiki/rendering/internal/renderer/xwiki20/XWikiSyntaxChainingRenderer.java
```diff
@@ -307,6 +307,8 @@ public void endFormat(Format format, Map<String, String> parameters)
                 break;
             case NONE:
                 break;
+            default : //none of the case matched
```
tmortagne: A more specific comment would be nicer :)

Like "Unsupported format" for example.

SrijanJha28: ok

SrijanJha28:
done

tmortagne:

PR #45, opened 2017-12-11
[codestyle]  replaced indexof(string) with indexof(char) for speed

ApCoder123:
resolves the issue here http://sonar.xwiki.org/issues/search#issues=5b92bb8b-b188-465e-9831-c9899b38c111


tmortagne:

PR #51, opened 2017-12-13
[codestyle] remove explicit init to default value

piyush69:


tmortagne:

PR #53, opened 2017-12-16
[codestyle] Replace the synchronized class "StringBuffer" by an unsynchronized one such as "StringBuilder"

SrijanJha28:


tmortagne:
Thanks for you contribution !

PR #54, opened 2017-12-18
[codestyle] Added default case to switch

TITANMQ:
One duplicated blocks of code removed.

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/MacroBlock.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/MacroBlock.java
@@ -51,7 +51,18 @@ public class MacroBlock extends AbstractBlock
      * The macro is located in a inline content (like paragraph, etc.).
      */
     private boolean inline;
-
+     
+     /**
+     * @param id the id of the macro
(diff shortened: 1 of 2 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/MacroBlock.java
```diff
@@ -52,16 +52,6 @@
      */
     private boolean inline;
 
-    /**
-     * @param id the id of the macro
-     * @param parameters the parameters of the macro
-     * @param isInline indicate if the macro is located in a inline content (like paragraph, etc.)
-     */
-    public MacroBlock(String id, Map<String, String> parameters, boolean isInline)
```
tmortagne: You don't explain what it's duplicating and deleting a public API is definitely a major breakage.

You should not blindly apply what you think sonar indicate and think about the consequences. As indicated in the GCI task, sometimes Sonar is wrong and you should ask on the chat when you are not sure.

tmortagne:
Was already fixed by another student.

Also please do you modification in a dedicated branch on your side and create a different pull request for each subject.

PR #55, opened 2017-12-18

SrijanJha28:
Updated WikiModelGeneratorListener.java by Adding a default case to this switch.


xwiki-rendering-syntaxes/xwiki-rendering-syntax-wikimodel/src/main/java/org/xwiki/rendering/internal/renderer/wikimodel/WikiModelGeneratorListener.java
```diff
@@ -149,6 +149,8 @@ public void beginFormat(Format format, Map<String, String> parameters)
             case NONE:
                 this.wikimodelListener.beginFormat(new WikiFormat(createWikiParameters(parameters).toList()));
                 break;
+            default: //Unsupported format
```
tmortagne: Would be more consistent if you could do the same for endFormat.

SrijanJha28:
Done :)

tmortagne:

PR #56, opened 2017-12-18
[codestyle] Variable 'lastEndFormat' explicitly initialized to 'null'  (default value for its type)

SrijanJha28:


tmortagne:

PR #60, opened 2017-12-18
Changed to use isEmpty

ApCoder123:
resolves issue here http://sonar.xwiki.org/issues/search#issues=0183e8e7-085f-44f3-9d71-c9f4bfe15454


tmortagne:

PR #61, opened 2017-12-18
Removed unused parameter

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=9153ccfb-c31c-491b-8280-4a4e1107481b

### DIFF
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/main/java/org/xwiki/rendering/internal/renderer/xwiki20/reference/XWikiSyntaxResourceRenderer.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/main/java/org/xwiki/rendering/internal/renderer/xwiki20/reference/XWikiSyntaxResourceRenderer.java
@@ -76,7 +76,7 @@ public class XWikiSyntaxResourceRenderer
         return result;
     }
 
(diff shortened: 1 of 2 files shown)


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/main/java/org/xwiki/rendering/internal/renderer/xwiki20/reference/XWikiSyntaxResourceRenderer.java
```diff
@@ -76,7 +76,7 @@ public String serialize(ResourceReference reference, boolean isFreeStanding)
         return result;
     }
 
-    public void beginRenderLink(XWikiSyntaxEscapeWikiPrinter printer, ResourceReference reference,
+    public void beginRenderLink(XWikiSyntaxEscapeWikiPrinter printer,
```
tmortagne: You also need to fix any code calling this method. You should always build the module you are modifying with Maven before committing. If you are using the XWiki dev VM just execute `mvn clean install -Pquality` in the parent folder containing the pom.xml file. If you don't use the VM see http://dev.xwiki.org/xwiki/bin/view/Community/Building.

ApCoder123:
I removed all usages of the removed parameter

tmortagne:
Thanks for your contribution.

PR #62, opened 2017-12-18
Added missing javadoc tag

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=a1a6ee90-dfc4-4295-9c4d-dac7ff79eb6f

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultTransformationManager.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultTransformationManager.java
@@ -79,10 +79,16 @@ public class DefaultTransformationManager implements TransformationManager
     @Named("context")
     private Provider<ComponentManager> componentManagerProvider;
 
-    @Override
     /**
-    * @deprecated
-    */
+     * {@inheritDoc}
+     * inherits on {@link TransformationManager}
+     *
+     * @deprecated
+     *
+     * Replaced by {@link #performTransformations(Block, TransformationContext)}
+     *
+     */
+    @Override
     @Deprecated
     public void performTransformations(XDOM dom, Syntax syntax) throws TransformationException
     {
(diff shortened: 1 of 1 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultTransformationManager.java
```diff
@@ -80,6 +80,9 @@
     private Provider<ComponentManager> componentManagerProvider;
 
     @Override
+    /**
+    * @deprecated
```
tmortagne: This javadoc is not correct:

* the javadoc is supposed to be locate before all the annotations
* you should indicate that this javadoc inherits the on in `TransformationManager`, see http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle#HDonotduplicateJavadoc. Any proper IDE should generate that for you (on Eclipse it's `Ctrl+Shift+j`)
* you should indicate what method should be used instead of this one

ApCoder123:
I've updated the Javadoc.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultTransformationManager.java
```diff
@@ -79,6 +79,15 @@
     @Named("context")
     private Provider<ComponentManager> componentManagerProvider;
 
+    /**
+     * {@inheritDoc}
+     * inherits on {@link TransformationManager}
```
tmortagne: You don't need this line.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultTransformationManager.java
```diff
@@ -79,6 +79,15 @@
...
+     *
+     * @deprecated
+     *
+     * Replaced by {@link #performTransformations(Block, TransformationContext)}
```
tmortagne: Should be located in the same line that `@deprecated` (this is the documentation of the `@deprecated` tag)

ApCoder123:
Fixed the issues

tmortagne:

PR #64, opened 2017-12-20
[codestyle]Reduced number of returns to 3

VQTran123:
Removed return statement for document reference because it seemed unnecessary and replaced it with return statement for resourceReference.


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -142,7 +142,7 @@ public ResourceReference parse(String rawReference)
         StringBuilder content = new StringBuilder(rawReference);
         resourceReference = parseInterWikiLinks(content);
         if (resourceReference != null) {
-            return resourceReference;
+            resourceReference=resourceReference;
```
tmortagne: * setting resourceReference in resourceReference is definitely going to irritate Sonar :)
* if isInWikiMode() return false then you won't get the same result than before

tmortagne:
Moved to https://github.com/xwiki/xwiki-rendering/pull/65

PR #65, opened 2017-12-20
[codestyle]Reduced number of returns to 3

VQTran123:
Removed step 4 as it seemed unnecessary and made step 3 an else statement so the code can compile.

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
@@ -146,14 +146,13 @@ public class GenericLinkReferenceParser extends AbstractResourceReferenceParser
         }
 
         // Step 3: If we're in non wiki mode, we consider the reference to be a URL.
-        if (!isInWikiMode()) {
+        else{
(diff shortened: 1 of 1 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -146,14 +146,13 @@ public ResourceReference parse(String rawReference)
...
-        if (!isInWikiMode()) {
+        else{
             resourceReference = new ResourceReference(rawReference, ResourceType.URL);
             resourceReference.setTyped(false);
             return resourceReference;
         }
 
-        // Step 4: Consider that we have a reference to a document
-        return parseDocumentLink(content);
+
```
tmortagne: What exactly makes you think it's unnecessary ? From what I understand step 4 is actually what happen most of the time (we almost never have interwiki links and in the context of a running XWiki instance you are on wiki mode).

Seems to me you are over thinking this. You are not supposed to modify the logic of the method, just improve the codestyle.

tmortagne:
Note: please try to reuse the same pull request (i.e. commit in the same branch) for the same subject.

tmortagne:
Please build the module with Maven before committing. If you did you would have seen this cannot work.

PR #66, opened 2017-12-20
Removed duplicated blocks of code

gonzaloUr:
Moved duplicated hashCode method in MacroBlock and MacroMakerBlock to a common class.

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
@@ -68,43 +68,29 @@ public abstract class AbstractMacroBlock extends AbstractBlock
     /**
      * @return the macro content.
      */
-    public String getContent() {
+    public String getContent()
+    {
         return content;
     }
 
     /**
      * @return the macro identifier.
      */
-    public String getId() {
+    public String getId()
+    {
         return id;
     }
 
     /**
      * @return if true the macro is located in a inline content (like paragraph, etc.).
      */
-    public boolean isInline() {
+    public boolean isInline()
+    {
         return inline;
     }
 
     /**
-     * Helper hashCode method for any AbstractMacroBlock.
-     *
-     * @return a hashcode.
-     */
-    protected int abstractBlockHashCode()
-    {   
-        HashCodeBuilder builder = new HashCodeBuilder();
-
-        builder.appendSuper(super.hashCode());
-        builder.append(getContent());
-        builder.append(getId());
-        builder.append(isInline());
-
-        return builder.toHashCode();
-    }
-
-    /**
-     * Helper equals method for any AbstractMacroBlock.
+     * Helper method for any AbstractMacroBlock equals method.
      *
      * @param obj AbstractMacroBlock.
      * @return true if obj is equals.
@@ -122,5 +108,24 @@ public abstract class AbstractMacroBlock extends AbstractBlock
         builder.append(isInline(), obj.isInline());
 
         return builder.isEquals();
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        return super.equals(obj);
+    }
+
+    @Override
+    public int hashCode()
+    {   
+        HashCodeBuilder builder = new HashCodeBuilder();
+
+        builder.appendSuper(super.hashCode());
+        builder.append(getContent());
+        builder.append(getId());
+        builder.append(isInline());
+
+        return builder.toHashCode();
     }
 }
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
@@ -1,0 +1,126 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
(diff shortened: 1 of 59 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Common class to MacroBlock and MacroMakerBlock.
+ *
+ * @version $Id$
+ * @since 9.11
+ */
+public abstract class AbstractMacroBlock extends AbstractBlock
+{
+
```
tmortagne: I doubt this is working since that would mean MacroBlock and MacroMakerBlock are calling a super constructor that does not exist.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
```diff
@@ -0,0 +1,43 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.rendering.block;
+
+/**
+ * Common class to MacroBlock and MacroMakerBlock.
+ *
+ * @version $Id$
+ * @since 9.11
```
tmortagne: 9.11 being out it now should be `10.0RC1`

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
```diff
@@ -0,0 +1,43 @@
...
+public abstract class AbstractMacroBlock extends AbstractBlock
+{
+
+    @Override
+    public int hashCode()
+    {	
+        HashCodeBuilder builder = new HashCodeBuilder();
+
+        builder.appendSuper(super.hashCode());
+        builder.append(getContent());
```
tmortagne: You are calling methods which don't exist at this level.

tmortagne:
AbstractMacroBlock is the right move but you are missing a lot of stuff in that class. Please don't commit anything until the build passes with Maven.

gonzaloUr:
I've already changed what was suggested and also added a unit test. My changes seem to build (including the test) but I'm not able to get the entire project. I've attached the full log as well as the parts where the unit test passes, because of this I'm pretty sure that the errors have nothing to do with my changes.

[log.txt](https://github.com/xwiki/xwiki-rendering/files/1581341/log.txt)
[log-test.txt](https://github.com/xwiki/xwiki-rendering/files/1581342/log-test.txt)

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public AbstractMacroBlock(List<? extends Block> childrenBlocks, Map<String, String> parameters)
+    {
+        super(childrenBlocks, parameters);
+    }
+
+    /**
+     * @return the macro content.
+     */
+    public abstract String getContent();
```
tmortagne: It does not make much sense to have all those abstract method. You should move the field since they are duplication too.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
```diff
@@ -0,0 +1,150 @@
...
+        super(childBlock, parameters);
+    }
+
+    /**
+     * @see AbstractMacroBlock#AbstractMacroBlock(List, Map)
+     *
+     * @param childrenBlocks the list of children blocks of the block to construct
+     * @param parameters the parameters to set
+     */
+    public AbstractMacroBlock(List<? extends Block> childrenBlocks, Map<String, String> parameters)
```
tmortagne: You don't need all those constructors. MacroBlock and MacroMarkerBlock only need one.

gonzaloUr:
I removed the constructors and abstract methods and moved the fields and getters to AbstractMacroBlock.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
```diff
@@ -0,0 +1,126 @@
...
+
+        this.id = id;
+        this.content = content;
+        this.inline = inline;
+    }
+
+    /**
+     * @return the macro content.
+     */
+    public String getContent() {
```
tmortagne: This does not follow XWiki Java codestyle (you are missing a newline, see the classes from which you copy/pasted that). See http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle#HConfiguringyourIDEtousetheXWikicodestyle for resource to setup Eclipse or IDEA formatters.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/MacroBlock.java
```diff
@@ -128,29 +83,16 @@ public boolean equals(Object obj)
...
+        if (!(obj instanceof MacroBlock)) {
+            return false;
         }
 
-        return false;
+        return abstractBlockEquals((MacroBlock) obj);
     }
 
     @Override
     public int hashCode()
```
tmortagne: Why not implementing hashCode at abstract level since there is nothing specific here.

gonzaloUr:
I had to override equals in AbstractMacroBlock and hashCode in both MacroBlock and MacroMakerBlock. It is not allowed by the checkstyle for a class to override equals but not hashCode and vice versa.

tmortagne:

PR #67, opened 2017-12-24
XRENDERING-61 : Box around TOC Macro

akshitdewan:
Am I on the right track with this? I just added the parameters and getters and setters. If I am on the right track, what's next! If not, what do I need to do?

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/pom.xml
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/pom.xml
@@ -42,5 +42,10 @@
       <version>4.3.0.Final</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-macro-box</artifactId>
+      <version>10.0-SNAPSHOT</version>
+    </dependency>
   </dependencies>
 </project>
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
@@ -34,6 +34,8 @@ import org.xwiki.rendering.block.XDOM;
 import org.xwiki.rendering.listener.reference.DocumentResourceReference;
 import org.xwiki.rendering.listener.reference.ResourceReference;
 import org.xwiki.rendering.macro.AbstractMacro;
+import org.xwiki.rendering.macro.Macro;
+import org.xwiki.rendering.macro.box.BoxMacroParameters;
 import org.xwiki.rendering.macro.MacroExecutionException;
 import org.xwiki.rendering.macro.toc.TocMacroParameters;
 import org.xwiki.rendering.parser.Parser;
@@ -57,7 +59,11 @@ public class TocMacro extends AbstractMacro<TocMacroParameters>
      * The description of the macro.
      */
     private static final String DESCRIPTION = "Generates a Table Of Contents.";
-
+    
+    @Inject
+    @Named("box")
+    private Macro<BoxMacroParameters> boxMacro;
+    
     private TocTreeBuilder tocTreeBuilder;
 
     /**
@@ -126,7 +132,9 @@ public class TocMacro extends AbstractMacro<TocMacroParameters>
         TreeParameters treeParameters = builder.build(rootBlock, parameters, context);
         List<Block> parametersList = this.tocTreeBuilder.build(treeParameters);
         if (parameters.getBox()) {
-            MacroBlock mb = new MacroBlock("toc", parametersList.get(0).getRoot().getParameters(), false);
+            BoxMacroParameters boxParameters = new BoxMacroParameters();
+            boxParameters.setCssClass(context.getId() + "message");
+            parametersList = boxMacro.execute(boxParameters, content, context);
         }
 
         return parametersList;
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
@@ -29,6 +29,7 @@ import javax.inject.Singleton;
 import org.xwiki.component.annotation.Component;
 import org.xwiki.component.phase.InitializationException;
 import org.xwiki.rendering.block.Block;
+import org.xwiki.rendering.block.MacroBlock;
 import org.xwiki.rendering.block.XDOM;
 import org.xwiki.rendering.listener.reference.DocumentResourceReference;
 import org.xwiki.rendering.listener.reference.ResourceReference;
@@ -123,7 +124,13 @@ public class TocMacro extends AbstractMacro<TocMacroParameters>
 
         TreeParametersBuilder builder = new TreeParametersBuilder();
         TreeParameters treeParameters = builder.build(rootBlock, parameters, context);
-        return this.tocTreeBuilder.build(treeParameters);
+        List<Block> parametersList = this.tocTreeBuilder.build(treeParameters);
+        if (parameters.getBox()) {
+            MacroBlock mb = new MacroBlock("toc", parametersList.get(0).getRoot().getParameters(), false);
+        }
+
+        return parametersList;
+
     }
 
     private XDOM getXDOM(ResourceReference reference, WikiModel wikiModel) throws MacroExecutionException
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
@@ -46,6 +46,22 @@ public class TocMacroParameters
          */
         PAGE
     }
+    
+    /**
+     * @version $Id$
+     */
+    public enum Floating
+    {
+        /**
+         * Floating to the left.
+         */
+        LEFT,
+        
+        /**
+         * Floating to the right.
+         */
+        RIGHT
+    }
 
     /**
      * The minimum section level. For example if 2 then level 1 sections will not be listed.
@@ -76,13 +92,16 @@ public class TocMacroParameters
     private boolean numbered;
     
     /**
-     * Set to 0 if floating to the left, 1 if floating to the right. Default : 1
+     * Whether TOC is floating to the left or the right.
      */
-    private int floating = 1;
-
+    private Floating floating = Floating.LEFT;
+    
     private String reference;
 
-    private boolean box = true;
+    /**
+     * Whether there is a box.
+     */
+    private boolean box;
 
     /**
      * @param start the minimum section level. For example if 2 then level 1 sections will not be listed.
@@ -185,26 +204,23 @@ public class TocMacroParameters
     }
     
     /**
-     * @param floating int of value 0 or 1 that sets the floating to left or right respectively
+     * @param floating Floating with value left or right
      */
-    public void setFloating(int floating)
+    public void setFloating(Floating floating)
     {
-        if (floating != 1 && floating != 0) {
-            return;
-        }
         this.floating = floating;
     }
 
     /**
-     * @return the int value of floating : 0 for left and 1 for right
+     * @return current value of floating
      */
-    public int getFloating()
+    public Floating getFloating()
     {
         return this.floating;
     }
 
     /**
-     * @param box whether or not the toc is a box
+     * @param box whether or not there is a box
      */
     public void setBox(boolean box)
     {
@@ -212,7 +228,7 @@ public class TocMacroParameters
     }
 
     /**
-     * @return whether or not the toc is a box
+     * @return whether or not there is a box
      */
     public boolean getBox()
     {
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
@@ -74,8 +74,15 @@ public class TocMacroParameters
      * If true the section title number is printed.
      */
     private boolean numbered;
+    
+    /**
+     * Set to 0 if floating to the left, 1 if floating to the right. Default : 1
+     */
+    private int floating = 1;
 
     private String reference;
+
+    private boolean box = true;
 
     /**
      * @param start the minimum section level. For example if 2 then level 1 sections will not be listed.
@@ -175,5 +182,40 @@ public class TocMacroParameters
     public String getReference()
     {
         return this.reference;
+    }
+    
+    /**
+     * @param floating int of value 0 or 1 that sets the floating to left or right respectively
+     */
+    public void setFloating(int floating)
+    {
+        if (floating != 1 && floating != 0) {
+            return;
+        }
+        this.floating = floating;
+    }
+
+    /**
+     * @return the int value of floating : 0 for left and 1 for right
+     */
+    public int getFloating()
+    {
+        return this.floating;
+    }
+
+    /**
+     * @param box whether or not the toc is a box
+     */
+    public void setBox(boolean box)
+    {
+        this.box = box;
+    }
+
+    /**
+     * @return whether or not the toc is a box
+     */
+    public boolean getBox()
+    {
+        return this.box;
     }
 }


vmassol:
Almost! As the issue says you should use "floating=left|right" rather than 0 or 1. Also by default there should be no box. Thanks :)

akshitdewan:
@vmassol what kind of variable would floating be, string with values left and right?
thanks!

vmassol:
Enum. See https://github.com/xwiki/xwiki-platform/blob/ade404691fcda3184abff5125c061bd71e2f0f41/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java#L36 for an example

akshitdewan:
@vmassol ohh that completely slipped my mind, thanks :)

akshitdewan:
@vmassol should be good now! I'm also not sure how this works, as I created the parameters but where are they being used?

vmassol:
Yes you need to use the parameters now in the macro code ;)

vmassol:
Check the macro's execute() method.

vmassol:
See http://rendering.xwiki.org/xwiki/bin/view/Main/ExtendingMacro for more details

akshitdewan:
okay, thanks!

akshitdewan:
@vmassol I'm a little confused, am I just setting the parameters of the parameters variable?

vmassol:
Your code is just allowing to set parameters for the macro. Now you need to use those parameter values so that the macro's execute() method returns the correct XDOM blocks.

akshitdewan:
Hi @vmassol , I looked at the link you sent and other macros, but I can't seem to figure out how I will "use those parameter values". All the macros use their parameters differently, and I'm not too familiar with how this all works. If it isn't of too much inconvenience to you, could you outline more specifically what I need to do?

vmassol:
There are several ways to implement this but I think the best is to reuse the Box Macro and thus if the "box" parameter is set to true for example, you should wrap the result of the TOC macro inside a MacroBlock. Hope it helps!

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
```diff
@@ -123,7 +124,13 @@ public boolean supportsInlineMode()
 
         TreeParametersBuilder builder = new TreeParametersBuilder();
         TreeParameters treeParameters = builder.build(rootBlock, parameters, context);
-        return this.tocTreeBuilder.build(treeParameters);
+        List<Block> parametersList = this.tocTreeBuilder.build(treeParameters);
+        if (parameters.getBox()) {
```
akshitdewan: Is this how I construct the MacroBlock? what do I do next?

akshitdewan:
@vmassol Did I construct the MacroBlock correctly?  I don't think I did :( what do I do after?

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
```diff
@@ -123,7 +124,13 @@ public boolean supportsInlineMode()
...
+            MacroBlock mb = new MacroBlock("toc", parametersList.get(0).getRoot().getParameters(), false);
```
tmortagne: I don't think this does what you expect. You are return a new toc macro here, not a box which seems to be the intent of this `if`.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
```diff
@@ -46,6 +46,22 @@
          */
         PAGE
     }
+    
+    /**
+     * @version $Id$
+     */
```
tmortagne: Don't forget to add a `@since 10.0RC1` in the javadoc for thi enum.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
```diff
@@ -176,4 +202,36 @@ public String getReference()
     {
         return this.reference;
     }
+    
+    /**
+     * @param floating Floating with value left or right
+     */
```
tmortagne: Same here, don't forget the `@since 10.0RC1`.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
```diff
@@ -176,4 +202,36 @@ public String getReference()
...
+    public void setFloating(Floating floating)
+    {
+        this.floating = floating;
+    }
+
+    /**
+     * @return current value of floating
+     */
```
tmortagne: Same here, don't forget the `@since 10.0RC1`.

Also you don't seems to be using this new parameter anywhere in the macro.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
```diff
@@ -176,4 +202,36 @@ public String getReference()
...
+    public Floating getFloating()
+    {
+        return this.floating;
+    }
+
+    /**
+     * @param box whether or not there is a box
+     */
```
tmortagne: Same here, don't forget the `@since 10.0RC1`.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
```diff
@@ -176,4 +202,36 @@ public String getReference()
...
+    public void setBox(boolean box)
+    {
+        this.box = box;
+    }
+
+    /**
+     * @return whether or not there is a box
+     */
```
tmortagne: Same here, don't forget the `@since 10.0RC1`.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
akshitdewan: @tmortagne I'm not sure how I would go about creating a box, do I create a BoxMacro object?

tmortagne: You can look at BoxMacro code to see how it's butting a box around the content.

vmassol: Yes but don't copy that. You need to reuse the Box Macro instead of copying it to avoid duplication and also to benefit from its future improvements.

akshitdewan:
For some reason, I am not able to import the box macro on the toc macro file on the local environment. Any ideas why? I'm able to import it from other macros but not here.

tmortagne:
If you want to reuse code from box macro module you need to make sure toc macro module has it as dependency in the pom.xml file.

xwiki-rendering-macros/xwiki-rendering-macro-toc/pom.xml
```diff
@@ -42,5 +42,10 @@
       <version>4.3.0.Final</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-macro-box</artifactId>
+      <version>10.0-SNAPSHOT</version>
```
vmassol: you should use `<version>${project.version}</version>` instead

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
```diff
@@ -123,7 +130,15 @@ public boolean supportsInlineMode()
 
         TreeParametersBuilder builder = new TreeParametersBuilder();
         TreeParameters treeParameters = builder.build(rootBlock, parameters, context);
-        return this.tocTreeBuilder.build(treeParameters);
+        List<Block> parametersList = this.tocTreeBuilder.build(treeParameters);
+        if (parameters.getBox()) {
+            BoxMacroParameters boxParameters = new BoxMacroParameters();
+            boxParameters.setCssClass(context.getId() + "message");
+            parametersList = boxMacro.execute(boxParameters, content, context);
```
akshitdewan: @tmortagne Is it possible to reuse the BoxMacro here? How would I go about setting the parameters?

xwiki-rendering-macros/xwiki-rendering-macro-toc/pom.xml
vmassol: and I think using `<optional>true</optional>` would be good too.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
```diff
@@ -123,7 +130,15 @@ public boolean supportsInlineMode()
 
         TreeParametersBuilder builder = new TreeParametersBuilder();
         TreeParameters treeParameters = builder.build(rootBlock, parameters, context);
-        return this.tocTreeBuilder.build(treeParameters);
+        List<Block> parametersList = this.tocTreeBuilder.build(treeParameters);
```
vmassol: The variable naming is a bit strange since this generates the list of Block for the output, they're not parameters. Proposed variable name: `macroBlocks` or `result`.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
vmassol: No you shouldn't execute the macro. Instead you should generate a `MacroBlock` and return it in the result.

vmassol: Note that this allows to declare the dependency in the pom.xml on the Box Macro as runtime only above.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
vmassol: And some javadoc :)

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocMacroParameters.java
```diff
@@ -74,9 +90,19 @@
...
+    
+    /**
+     * Whether TOC is floating to the left or the right.
+     */
+    private Floating floating = Floating.LEFT;
+    
     private String reference;
 
+    /**
+     * Whether there is a box.
```
vmassol: --> "Whether there is a box around or the TOC or not"

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocMacro.java
akshitdewan: How would I generate a MacroBlock while using the box macro?

vmassol: `new MacroBlock("box", ...)`. See the javadoc for MacroBlock to see the parameters you need to pass.

tmortagne: I'm really not a fan of the MacroBlock idea since it means you need to serialize a toc macro in wiki syntax (without the box parameter this time) and put it as macro block content.

vmassol: hmm indeed hadn't thought of the fact that MacroBlock only supports String is an issue... Maybe an idea would be to make MacroBlock support List<Block>. However this would also mean adding a new Macro#execute() signature. Could be interesting for the future.

But I agree it's not the right approach for now. I'd put a put comment in the code explaining why we cannot use a MacroBlock for the moment.

akshitdewan: If I shouldn't use the MacroBlock, what should I explore?

vmassol: I was going to say calling execute() but the problem is the same actually since it expects a String... :) @tmortagne what did you have in mind?

If the only solution is to not reuse the Box Macro then I think I prefer serialization rather than code duplication.

tmortagne: You could put the actual "boxing" code in some place where toc macro can reuse it (like a simple static public method in BoxMacro for example or a cleaner component used by both box and code macro).

vmassol: Yes but this requires refactoring and I don't really like this way of reusing. It would be much nicer to see the box macro as a black box and not have too strong dependencies between macros. I really prefer a runtime dep.

vmassol: If we really don't want the serialization then I prefer to close this issue as won't fix since users can still manually use the box macro around the toc macro.

tmortagne: It's not really a big refactoring. Having to serialize the content is a big -1 from me, macros should only manipulate blocks.

vmassol: ok and I am -1 to introduce a strong build time dependency between the TOC macro and the Box Macro just for an optional parameter especially since the user should be able to write:

```
{{box...}}
{toc/}}
{{/box}}
```

So I guess this makes the issue a won't fix FTM?

tmortagne: If toc and box macro cannot share code the toc macro could have its own stuff. We are only talking about a GroupBlock with a specific class here so it's not a big copy/paste...

akshitdewan: Do we plan to fix this issue?

tmortagne: IMO should should simply implement a box around the tox content directly (without reusing BoxMacro code), it's not like it was a lot of code. When I told you to look at BoxMacro code it was to see how to create a box using a block.

vmassol: I still need to think about it more but I don't like it too much ATM and I'm no longer sure we need this jira issue. I'm leaning towards recommending to use the box macro around the toc macro when needing a box. Same as using the cache macro around other macros when you need a cache.

akshitdewan: looks like it uses a boxblockbuilder class, should I look into that?

tmortagne: Yes that's where the code seems to be located but BoxBlockBuilder handle a lot more use cases that you (toc is never inline, no image reference, no title, etc.) o you need only a fraction of what BoxBlockBuilder (only the GroupBlock with the right css classes in it basically).

PR #68, opened 2017-12-25
Avoid declaring a variable if it is unreferenced before a possible exit point

RohitTanikella:
I moved the declaration of the variable so it would not be premature.

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/XDOMBuilder.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/XDOMBuilder.java
@@ -52,13 +52,14 @@ public class XDOMBuilder
      */
     public XDOM getXDOM()
     {
-        List<Block> blocks = endBlockList();
-
+        
         if (!this.stack.isEmpty()) {
             throw new IllegalStateException("Unbalanced begin/end Block events, missing " + this.stack.size()
                 + " calls to endBlockList().");
         }
-
+        
+        List<Block> blocks = endBlockList();
+        
         // support even events without begin/endDocument for partial content
         if (!blocks.isEmpty() && blocks.get(0) instanceof XDOM) {
             return (XDOM) blocks.get(0);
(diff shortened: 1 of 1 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/XDOMBuilder.java
```diff
@@ -52,13 +52,14 @@ public XDOMBuilder()
...
-        List<Block> blocks = endBlockList();
-
+        
         if (!this.stack.isEmpty()) {
             throw new IllegalStateException("Unbalanced begin/end Block events, missing " + this.stack.size()
                 + " calls to endBlockList().");
         }
-
+        
+        List<Block> blocks = endBlockList();
```
vmassol: hmm this is not going to have the same effect since `endBlockList()` calls `this.stack.pop()` , and thus `this.stack.isEmpty()` is going to have a different result value.

RohitTanikella: So would this be a false positive or is there another way to fix this?

RohitTanikella: I think I found a solution can you check it?

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/XDOMBuilder.java
```diff
@@ -52,13 +52,14 @@ public XDOMBuilder()
      */
     public XDOM getXDOM()
     {
-        List<Block> blocks = endBlockList();
-
-        if (!this.stack.isEmpty()) {
+        
+        if (!this.stack.size()==1) {
```
tmortagne: This is not exactly the same meaning as previous code. You should check if the size is > 1.

tmortagne:
This is a false positive. I closed the Sonar issue indicating it was found by you.

PR #73, opened 2017-12-25
[codestyle] use char instead of length 1 string

akshitdewan:
http://sonar.xwiki.org/issues/search#issues=befedb46-5df3-4bec-adc8-b1f60e6839ea


tmortagne:

PR #75, opened 2017-12-26
Made variable private and added accessors

ApCoder123:
Made a variable private and added accessors and changed usages in other classes. Resolves issue here: http://sonar.xwiki.org/issues/search#issues=5640c6a6-c95c-42a3-9dee-6a7a52fe5026


tmortagne:
Same as #81

PR #76, opened 2017-12-26
Made variable private

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=98fd67de-a03a-4367-8da0-e735c5dd7501

No accessors are needed as the variable is only used inside that class


tmortagne:
Same as #81

PR #77, opened 2017-12-26
Made variable private. No accessors needed

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=c99b2f82-ff20-4e79-b012-1fda5ab5c7d4


tmortagne:
Same as #81

PR #78, opened 2017-12-26
Made variable `prefix` private & added accessors

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=44a0435a-235c-471d-84e0-467ce7a9cb9f


tmortagne:
Same as #81

PR #79, opened 2017-12-26
Made syntaxExtension private & added accessors

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=5932c4bd-7ece-4ec5-b6a0-461ad51a7401


tmortagne:
Same as #81

PR #80, opened 2017-12-26
Made ctsExtension private & added accessors

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=7bd34acf-fa80-44c9-9b90-e238a56c9650


tmortagne:
Same as #81.

PR #81, opened 2017-12-26
Made state private & added accessors

ApCoder123:
Resolves issue here: http://sonar.xwiki.org/issues/search#issues=d801487a-ca8f-4587-a576-0542e87fc11c


tmortagne:
This is really not required. Also your change actually makes things worst because you end up with one field non consistent with others.

tmortagne:
I closed the Sonar issue and indicated it has been found by you.

PR #85, opened 2017-12-27
[codestyle] removed redundant method

AsianMercenary:
*Added !(obj instanceof AbstractMacroBlock) in case obj parameter is an instance of Block (addressed in super.equals(obj)) but not an AbstractMacroBlock
*Deleted private equals(AbstractMacroBlock obj) as it is solely called in the public equals(Object obj)

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
@@ -95,21 +95,17 @@ public abstract class AbstractMacroBlock extends AbstractBlock
         if (this == obj) {
             return true;
         }
-
-        if (!super.equals(obj)) {
+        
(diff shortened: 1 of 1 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractMacroBlock.java
```diff
@@ -95,21 +95,17 @@ public boolean equals(Object obj)
         if (this == obj) {
             return true;
         }
-
-        if (!super.equals(obj)) {
+        
+        if (!super.equals(obj) || !(obj instanceof AbstractMacroBlock)) {
```
gonzaloUr: Wouldn't that make equals not symmetric. For example, if you were going to create a new subclass of AbstractMacroBlock that required equals to check if some fields are the same or not.(https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)).

tmortagne:
This is actually a false positive.

PR #89, opened 2017-12-27

SrijanJha28:
1. http://sonar.xwiki.org/issues/search#issues=c306549e-52b5-484b-8ff6-2bfd8aa26764
2. http://sonar.xwiki.org/issues/search#issues=22676831-efd1-4a58-80a1-8897e32e7810


vmassol:
I'm not sure about this one. I think that if the format is not supported then the next "if" shouldn't be executed either. But let's consider this an improvement and a different issue. What you did isn't worse than what was there before ;)

PR #90, opened 2017-12-27
[codestyle] block of commented-out lines of code should be removed

SrijanJha28:


vmassol:
This looks definitely wrong. Why remove some comment?

vmassol:
Looks like a false positive

PR #93, opened 2017-12-27

SrijanJha28:


xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/reference/ResourceReference.java
```diff
@@ -283,15 +283,12 @@ public int hashCode()
     @Override
     public boolean equals(Object object)
     {
-        if (object == null) {
+        if (!(object instanceof ResourceReference)) {
```
vmassol: This is changing the logic and looks wrong. Where is null checked?

SrijanJha28: But @tmortagne had approved this in this pr : https://github.com/xwiki/xwiki-rendering/pull/40

SrijanJha28: and he only suggested me to do this.

vmassol: Have you tested it by passing "null" to see if it works?

vmassol: You can easily do that in the existing ResourceReferenceTest. Does it pass BTW? Note that it's missing a null check in the test.

SrijanJha28: ya it works

vmassol: FYI I've just improved the test: https://github.com/xwiki/xwiki-rendering/commit/738ebb69402ca37cb29cb586cb0d2eeaf57fe74e

vmassol: Just tested it too and indeed it works. "object instanceof ResourceReference" returns false when object is null.

vmassol: I find the new code less readable than the old one, even if it's more concise. I'd personally not have changed it but let's say it's ok...

SrijanJha28: I told you :)

vmassol: First rule of coding: never believe someone, always check ;)

PR #100, opened 2017-12-29
[codestyle] Reduce this switch case number of lines from 9 to at most 5, for example by extracting code into methods.

fennecus-z:
This is my second issue fix, the issue being located here:

http://sonar.xwiki.org/issues/search#issues=6b404718-9854-4136-9cae-dd35b87cb768

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/match/BlockNavigator.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/match/BlockNavigator.java
@@ -204,6 +204,14 @@ public class BlockNavigator
         return newBlocks;
     }
 
+    private Axes returnFollowingAxes(Block block, Axes currentAxes) {
+        if (!block.getChildren().isEmpty()) {
+            return Axes.FOLLOWING_SIBLING;
+        }
+
+        return currentAxes;
+    }
+
     // First block
 
     /**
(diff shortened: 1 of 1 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/match/BlockNavigator.java
```diff
@@ -239,15 +239,15 @@ public BlockNavigator(BlockMatcher matcher)
...
-                    List<Block> children = block.getChildren();
-                    if (!children.isEmpty()) {
-                        nextBlock = children.get(0);
+                    if (!block.getChildren().isEmpty()) {
                         axes = Axes.FOLLOWING_SIBLING;
-                        if (this.matcher.match(nextBlock)) {
-                            return (T) nextBlock;
-                        }
-                    }
+                        if (this.matcher.match(nextBlock)) { return (T) block.getChildren().get(0); } }
```
tmortagne: Just removing the newlines is cheating :)

This does not follow XWiki code style so you are changing a violation for another here.

fennecus-z:
I've made appropriate changes. Kindly review. Thanks!

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/match/BlockNavigator.java
```diff
@@ -239,15 +247,22 @@ public BlockNavigator(BlockMatcher matcher)
...
+                   The function returns Axes.FOLLOWING_SIBLING if the condition is true, and if false returns
+                   the current axes.
+                 */
                 case CHILD:
-                    List<Block> children = block.getChildren();
-                    if (!children.isEmpty()) {
-                        nextBlock = children.get(0);
-                        axes = Axes.FOLLOWING_SIBLING;
+                    axes = setAxesToFollowing(block, currentAxes);
                         if (this.matcher.match(nextBlock)) {
```
tmortagne: The indentation seems wrong here.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/match/BlockNavigator.java
```diff
@@ -204,6 +204,14 @@ public BlockNavigator(BlockMatcher matcher)
         return newBlocks;
     }
 
+    private Axes setAxesToFollowing(Block block, Axes currentAxes) {
```
tmortagne: Despite its name this method does not seems to **set** anything.

fennecus-z:
I've made appropriate corrections

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/match/BlockNavigator.java
```diff
@@ -239,14 +247,21 @@ public BlockNavigator(BlockMatcher matcher)
...
+                        children = block.getChildren()
+                   The function returnFollowingAxes executes takes the place of the previously used code, shown below
+                        if (!block.getChildren().isEmpty()) {
+                            axes = Axes.FOLLOWING_SIBLING;
                         }
+                   The function returns Axes.FOLLOWING_SIBLING if the condition is true, and if false returns
+                   the current axes.
+                 */
+                case CHILD:
+                    axes = returnFollowingAxes(block, currentAxes);
```
tmortagne: I don't like the fact that this block become inconsistent with others (return a block instead of setting it).

fennecus-z: What do you suggest I do?

tmortagne: Not idea but right now I still prefer the previous version.

PR #101, opened 2017-12-29
[Codestyle] reduce returns

fennecus-z:
Hello, for my third and final issue, I decided to do this one

http://sonar.xwiki.org/issues/search#issues=c7e5a5f5-4a68-46d2-af23-08f3f7156ac4

Thanks, will be submitting my work. The link to my task is

https://codein.withgoogle.com/dashboard/task-instances/5241134932557824/

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
@@ -129,31 +129,45 @@ public class GenericLinkReferenceParser extends AbstractResourceReferenceParser
         return URI_PREFIXES;
     }
 
+
+    /*
+    Changes:
+    I set up a nested if statement to check if each value of resourceReference is null at each step.
+    If it is, it moves on to the next step and performs an action, then checks to see if it is still null.
+    If at the end of this all, the value of resourceReference is not null, resourceReference will be returned.
+    If it is still null, it returns a reference to a document.
+    This halves the number of return statement from 4 to 2, while only parsing when required.
+     */
     @Override
     public ResourceReference parse(String rawReference)
     {
         // Step 1: Check if it's a known URI by looking for one of the known URI schemes. If not, check if it's a URL.
         ResourceReference resourceReference = parseURILinks(rawReference);
-        if (resourceReference != null) {
-            return resourceReference;
-        }
-
-        // Step 2: Look for an InterWiki link
         StringBuilder content = new StringBuilder(rawReference);
-        resourceReference = parseInterWikiLinks(content);
+
+        if (resourceReference == null) {
+
+            // Step 2: Look for an InterWiki link
+            resourceReference = parseInterWikiLinks(content);
+            if (resourceReference == null) {
+
+                // Step 3: If we're in non wiki mode, we consider the reference to be a URL.
+                if (!isInWikiMode()) {
+                    resourceReference = new ResourceReference(rawReference, ResourceType.URL);
+                    resourceReference.setTyped(false);
+                }
+
+            }
+
+        }
+
         if (resourceReference != null) {
             return resourceReference;
+            // Step 4: Consider that we have a reference to a document
+            else {
+                return parseDocumentLink(content);
+            }
         }
-
-        // Step 3: If we're in non wiki mode, we consider the reference to be a URL.
-        if (!isInWikiMode()) {
-            resourceReference = new ResourceReference(rawReference, ResourceType.URL);
-            resourceReference.setTyped(false);
-            return resourceReference;
-        }
-
-        // Step 4: Consider that we have a reference to a document
-        return parseDocumentLink(content);
     }
 
     /**


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -132,28 +132,28 @@
...
+        if (parseURILinks(rawReference) != null) {
+            resourceReference = parseURILinks(rawReference);
+        }
         // Step 2: Look for an InterWiki link
-        StringBuilder content = new StringBuilder(rawReference);
-        resourceReference = parseInterWikiLinks(content);
-        if (resourceReference != null) {
-            return resourceReference;
+        else if (parseInterWikiLinks(content) != null) {
+            resourceReference = parseInterWikiLinks(content);
```
tmortagne: This is very bad for performances, you are re-executing several time the same code.

fennecus-z: Didn’t see that, will fix. Thanks.

tmortagne: You actually did this mistake in most of your modifications (parseURILinks is now executed 3 times now).

fennecus-z: As of the latest commit, the repeated executions should all be fixed. Thanks!

fennecus-z:
I've made the changes.

tmortagne:
You need to understand that doing useless parsing is costly. If you really need to touch this code (probably not a good idea IMO) you need to end up with something that is not slower that what you are "fixing".

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -132,28 +132,30 @@
     @Override
     public ResourceReference parse(String rawReference)
     {
+        StringBuilder content = new StringBuilder(rawReference);
+        ResourceReference parsedURILinks = parseURILinks(rawReference);
+        ResourceReference parsedInterWikiLinks = parseInterWikiLinks(content);
```
tmortagne: It does not make sense to do this before knowing if you will need it in the first place.

fennecus-z:
I've made another set of changes that does not perform unnecessary parsing, while reducing the number of returns to two

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -129,31 +129,45 @@
...
+    /*
+    Changes:
+    I set up a nested if statement to check if each value of resourceReference is null at each step.
+    If it is, it moves on to the next step and performs an action, then checks to see if it is still null.
+    If at the end of this all, the value of resourceReference is not null, resourceReference will be returned.
+    If it is still null, it returns a reference to a document.
+    This halves the number of return statement from 4 to 2, while only parsing when required.
+     */
     @Override
     public ResourceReference parse(String rawReference)
```
tmortagne: I still find the whole thing harder to read than the previous version which had very clear and separated steps. Keep in mind that having several returns is not a Java bug, it's just a readability advice from Sonar so it should not be fixed at all cost: only if you can find a more readable version which is not the case here.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -129,31 +129,45 @@
...
     {
         // Step 1: Check if it's a known URI by looking for one of the known URI schemes. If not, check if it's a URL.
         ResourceReference resourceReference = parseURILinks(rawReference);
-        if (resourceReference != null) {
-            return resourceReference;
-        }
-
-        // Step 2: Look for an InterWiki link
         StringBuilder content = new StringBuilder(rawReference);
```
tmortagne: This new StringBuilder might not always be needed.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -129,31 +129,45 @@
...
+
         }
 
-        // Step 3: If we're in non wiki mode, we consider the reference to be a URL.
-        if (!isInWikiMode()) {
-            resourceReference = new ResourceReference(rawReference, ResourceType.URL);
-            resourceReference.setTyped(false);
+        if (resourceReference != null) {
             return resourceReference;
+            // Step 4: Consider that we have a reference to a document
```
tmortagne: This comment would make more sense if put in the `else`.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -129,31 +129,45 @@
...
+            else {
```
tmortagne: You don't need a `else` when you have a return in the `if`. It would be a lot more readable if you keep the `Step 4` separated.

fennecus-z: Will make the correction

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
fennecus-z: Will make the correction

PR #109, opened 2017-12-31
3 Code Violations Fixed

RohitTanikella:
http://sonar.xwiki.org/issues/search#issues=85dbcbc5-e03a-4a4a-a81c-16bf6fbb92d6
http://sonar.xwiki.org/issues/search#issues=a6a2701c-f668-4392-9fa6-13dfdf251466
http://sonar.xwiki.org/issues/search#issues=8d8a73a3-470e-4c8e-b719-9924a73b4d27

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractBlock.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractBlock.java
@@ -547,7 +547,7 @@ public abstract class AbstractBlock implements Block
      * @param listener the listener that will receive the events sent by this block before its children blocks have
      *            emitted their own events.
      */
-    public void after(Listener listener)
+    public abstract void after(Listener listener)
     {
         // Do nothing by default, should be overridden by extending Blocks
     }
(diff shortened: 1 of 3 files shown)


RohitTanikella:
I did not claim the task yet because I can not while waiting for my false positive tasks to be approved.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractBlock.java
```diff
@@ -547,7 +547,7 @@ public void before(Listener listener)
      * @param listener the listener that will receive the events sent by this block before its children blocks have
      *            emitted their own events.
      */
-    public void after(Listener listener)
+    public abstract void after(Listener listener)
```
tmortagne: This is a false positive.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/GenericLinkReferenceParser.java
```diff
@@ -263,19 +263,16 @@ protected String parseElementAfterString(StringBuilder content, String separator
 
         // Find the first non escaped separator (starting from the end of the content buffer).
         int index = content.lastIndexOf(separator);
-        while (index != -1) {
+        while (index > 0) {
```
tmortagne: This will break the use case where the separator is the first character of the string.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/macro/DefaultMacroManager.java
```diff
@@ -106,6 +104,9 @@
                     + "For example \"html/xwiki/2.0\". This macro will not be available in the system.");
                 continue;
             }
+            
+            Set<MacroId> result = new HashSet<MacroId>();
```
tmortagne: Please build with Maven before committing. Also even if this was valid Java it would break the whole method (look at what this method is supposed to do).

PR #110, opened 2018-01-02
[codestyle]Rename booleans "getX" to "isX"

VQTran123:
Changed the names of "getWiki()" and "getClean()" to "isWiki()" and "isClean()".


vmassol:
Have you tested your change by running the build? This is going to generate a backward compatibility breakage from Revapi. I'm also not sure that the Macro engine support `is*()` vs `get*()`.

tmortagne:
`is*()` works fine but that's not the main issue.

HTMLMacroParameters is used from Java so you can't do that. IMO this is a won't fix, too late.

PR #114, opened 2018-01-04
Update EventType.java - using varargs

siddhpant:
Issue: Consider using varargs for methods or constructors which take an array the last parameter. (x52)

One permalink: http://sonar.xwiki.org/issues/search#issues=a104439b-a2f0-4abe-87a5-bd132d3b2564

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/QueueListener.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/QueueListener.java
@@ -54,7 +54,7 @@ public class QueueListener extends LinkedList<QueueListener.Event> implements Li
         /**
          * The parameters of the event.
          */
-        public Object... eventParameters;
+        public Object[] eventParameters;
 
         /**
          * @param eventType the type of the event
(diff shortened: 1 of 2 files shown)


tmortagne:
Ideally you should also update the code that use those methods to pass vararg instead of creating an array.

siddhpant:
vararg is backward compatible, I had a chat on the IRC.

04/01/2018 6:55 pm को "Thomas Mortagne" <notifications@github.com> ने लिखा:

tmortagne:
I know it's backward compatible, I did not suggested that you broken any API. I'm just saying that there is code in the same repository which could use those new APIs better now and would be nice to update it.

siddhpant:
Okay, I found two files which import EventType.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/QueueListener.java
```diff
@@ -54,13 +54,13 @@
         /**
          * The parameters of the event.
          */
-        public Object[] eventParameters;
+        public Object... eventParameters;
```
tmortagne: This is not valid Java.

tmortagne:
Looks like all use go trough Event constructor so fixing it should be enough.

siddhpant:
My bad. I was in a middle of something and also I was on mobile. Next time I would take care.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/QueueListener.java
siddhpant: Man that's too obvious.

siddhpant:
Done.

tmortagne:

PR #116, opened 2018-01-04
Fixed 3 coding violations

siddhpant:
Update DefaultMacroManager.java
Fixed http://sonar.xwiki.org/issues/search#issues=a6a2701c-f668-4392-9fa6-13dfdf251466

Update DoxiaPrinterAdapter.java
Fixed http://sonar.xwiki.org/issues/search#issues=6efc3f5c-99f8-4abb-bb3c-fabefba023c2

Update MacroId.java
Fixed http://sonar.xwiki.org/issues/search#issues=b79f65b8-dea8-47d6-ada6-7921845ae614

### DIFF
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/MacroId.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/MacroId.java
@@ -106,14 +106,11 @@ public class MacroId
     @Override
     public boolean equals(Object object)
     {
-        if (object == null) {
+        if ((object == null) || (object.getClass() != getClass())) {
             return false;
         }
         if (object == this) {
             return true;
-        }
-        if (object.getClass() != getClass()) {
-            return false;
         }
         MacroId rhs = (MacroId) object;
         return new EqualsBuilder()
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/MacroId.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/MacroId.java
@@ -106,12 +106,14 @@ public class MacroId
     @Override
     public boolean equals(Object object)
     {
-        if (object == null || object.getClass() != getClass()) {
-            return false;
-        }
         if (object == this) {
             return true;
         }
+        
+        if (!(object instanceof MacroId)) {             // Can't be null if instance of MacroId.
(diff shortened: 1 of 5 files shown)


xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/MacroId.java
```diff
@@ -106,15 +106,12 @@ public int hashCode()
     @Override
     public boolean equals(Object object)
     {
-        if (object == null) {
+        if ((object == null) || (object.getClass() != getClass())) {
```
tmortagne: No need for those parenthesis (they will even produce a Sonar issue).

siddhpant:
Done

05/01/2018 13:46 को "Thomas Mortagne" <notifications@github.com> ने लिखा:

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/MacroId.java
```diff
@@ -106,15 +106,12 @@ public int hashCode()
     @Override
     public boolean equals(Object object)
     {
-        if (object == null) {
+        if (object == null || object.getClass() != getClass()) {
```
tmortagne: Actually since MacroId is not really supposed to be extended you could replace this with a `if (!(object instanceof MacroId))`. Can't be null if it's an instance of MacroId.

Also I would put it after the `if (object == this)` test instead of before.

siddhpant: Done

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/MacroId.java
```diff
@@ -106,15 +106,14 @@ public int hashCode()
...
-        if (object == null) {
-            return false;
-        }
         if (object == this) {
             return true;
         }
-        if (object.getClass() != getClass()) {
-            return false;
+        
+        if (!(object instanceof MacroId)) {             // Can't be null if instance of MacroId.
```
tmortagne: I don't think those comments are really required. If you absolutely want to put them put them before the if and not on the same line.

siddhpant: Okay removed them

tmortagne:

PR #117, opened 2018-01-05
Fixed some coding violations

siddhpant:
Check the recent 5 commits (After merging of branch).


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiXHTMLWhitespaceXMLFilter.java
```diff
@@ -78,8 +78,8 @@ public void endCDATA() throws SAXException
     {
         if (getContent().length() > 0 && this.containsWikiSyntax) {
                 // Make sure we clean head/trail white spaces
```
tmortagne: Should also think the indentation here.

siddhpant: Done

PR #124, opened 2018-01-09
Fix some DefaultRenderingContext issues

gonzaloUr:
issues:

http://sonar.xwiki.org/issues/search#issues=4ed3cb35-b99a-4f07-8d75-63cec2a78d6f
http://sonar.xwiki.org/issues/search#issues=32df68df-dc42-47f7-ba29-dbfe94168e99

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultRenderingContext.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultRenderingContext.java
@@ -22,6 +22,7 @@ package org.xwiki.rendering.internal.transformation;
 
 import java.util.ArrayDeque;
 import java.util.Deque;
+import java.util.Collections;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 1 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultRenderingContext.java
```diff
@@ -202,7 +203,7 @@ public void transformInContext(Transformation transformation, TransformationCont
             return stack;
         }
 
-        return null;
+        return Collections.emptyList();
```
tmortagne: As you can see from the using it  it's expected that this method might return null. Also this code won't build.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultRenderingContext.java
```diff
@@ -266,7 +267,7 @@ public Syntax getTargetSyntax()
     public void setTargetSyntax(Syntax targetSyntax)
     {
         Context context = peek();
-        if (context != null && context != NULL_CONTEXT) {
+        if (context != null && context.equals(NULL_CONTEXT)) {
```
tmortagne: This was done on purpose. NULL_CONTEXT is a special instance here.

tmortagne:
False positives.

PR #130, opened 2018-01-10
[codestyle] declare parameter 'headerBlock' as org.xwiki.rendering.block.Block

fennecus-z:
This is a fix for the issue at http://sonar.xwiki.org/issues/search#issues=930a5c9b-5704-41e7-befd-81484244699f.

My GCI task is:

https://codein.withgoogle.com/dashboard/task-instances/5241134932557824/


tmortagne:
Toc job is to work with HeaderBlock so I prefer to keep it like this (the implementation of this method might change in a way that only works with HeaderBlock).

PR #143, opened 2018-09-26

surli:
* Create the proper MetaDataBlock
  * Add the type of content in ContentDescriptor
  * Change following macro to use new features: Box, Content, Figures, Message


surli:
@vmassol @tmortagne @mflorea could you review this? This is the first changes for the proposal, I used the last emails sent by Vincent to make those: I'm waiting for your comments before starting to work on WikiMacro. Thanks.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/ContentDescriptor.java
```diff
@@ -36,4 +43,16 @@
...
+    /**
+     * This method will return the type of the macro content. By default it fallback to List&lt;Block&gt;.
+     *
+     * @return the type of the macro content.
+     * @since 10.9-rc-1
+     */
+    @Unstable
+    default Type getType()
+    {
+        return new DefaultParameterizedType(null, List.class, Block.class);
```
mflorea: The default type should be ``String.class`` I think.

xwiki-rendering-macros/xwiki-rendering-macro-content/src/main/java/org/xwiki/rendering/internal/macro/content/SyntaxMacro.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * The syntax used by {@link ContentMacro} for the inline edition.
+ *
+ * @version $Id$
+ * @since 10.9-rc-1
+ */
+@Role
+@Named("content")
+public class SyntaxMacro implements SyntaxDescriptor
```
mflorea: The name ``SyntaxMacro`` sounds strange. It doesn't match the description.

xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/internal/macro/box/DefaultBoxMacro.java
```diff
@@ -59,7 +62,8 @@
      */
     public DefaultBoxMacro()
     {
-        super("Box", DESCRIPTION, new DefaultContentDescriptor(CONTENT_DESCRIPTION), BoxMacroParameters.class);
+        super("Box", DESCRIPTION, new DefaultContentDescriptor(CONTENT_DESCRIPTION, true,
+            new DefaultParameterizedType(null, List.class, Block.class)), BoxMacroParameters.class);
```
tmortagne: Since `new DefaultParameterizedType(null, List.class, Block.class)` is a very common use case you should probably put it in a constant somewhere. Probably in Block.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/InlineEditableContentBlock.java
```diff
@@ -0,0 +1,43 @@
...
+@Unstable
+public class InlineEditableContentBlock extends MetaDataBlock
+{
+    /**
+     * @param childBlocks the list of children blocks of the block to construct
+     * @see AbstractBlock#AbstractBlock(List)
+     */
+    public InlineEditableContentBlock(List<? extends Block> childBlocks)
+    {
+        super(childBlocks);
```
tmortagne: Isn't the point of this class to set some metadata ?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/ContentDescriptor.java
tmortagne: Yes since previous behavior correspond to String type the default should be String.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/SyntaxDescriptor.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Unstable
+@Role
+public interface SyntaxDescriptor
+{
+    /**
+     * @param macroParameters the parameters of the macro to be used.
+     * @return the syntax of the macro computed from information given in its parameters.
+     */
+    Syntax getSyntax(Map<String, Object> macroParameters);
```
tmortagne: How is the WYSIWYG supposed to use that ?

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/InlineEditableContentBlock.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.List;
+
+import org.xwiki.stability.Unstable;
+
+/**
+ * This block represents a part of a macro content which can be edited inline.
+ *
+ * @version $Id$
+ * @since 10.9-rc-1
```
atallahade: Should be 10.9RC1

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/SyntaxDescriptor.java
surli: AFAI understood @vmassol proposal the idea is to retrieve the component with the hint (see https://github.com/xwiki/xwiki-rendering/pull/143/files/11802d7856f36c09831c07a2231e396c082d29cf#diff-01939d2b6482e4ca74d1e171539afaad) and I assume the WYSIWYG already have the map of parameters with their value: then the component is able to determine the syntax and to return it. 

I put this proposal on purpose in the PR to be able to discuss it, but I also can remove it and put this in another PR to avoid mixing things: as @vmassol stated it could be a step 2.

tmortagne: > but I also can remove it and put this in another PR to avoid mixing things

Would be better yes.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/Block.java
```diff
@@ -35,6 +37,11 @@
  */
 public interface Block extends Cloneable
 {
+    /**
+     * Helper to represent the Type for each List&lt;Block&gt;.
+     */
+    Type LIST_BLOCK_TYPE = new DefaultParameterizedType(null, List.class, Block.class);
```
tmortagne: Will need a `@since`.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/InlineEditableContentBlock.java
```diff
@@ -0,0 +1,70 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * This block represents a part of a macro content which can be edited inline.
+ *
+ * @version $Id$
+ * @since 10.9RC1
+ */
+@Unstable
+public class InlineEditableContentBlock extends MetaDataBlock
```
tmortagne: I feel that the naming and documentation should reflect something a little less WYSIWYG specific: the fact that the source of this block is not modified during the macro execution.

xwiki-rendering-macros/xwiki-rendering-macro-box/src/test/resources/macrobox1.test
```diff
@@ -13,6 +13,7 @@ beginDocument
 beginMacroMarkerStandalone [box] [] [* one
 * two]
 beginGroup [[class]=[box]]
+beginMetaData
```
tmortagne: I guess you just forgot to update the tests after adding the metadata value to InlineEditableContentBlock ?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/SyntaxDescriptor.java
surli: OK, I'll do that then

xwiki-rendering-macros/xwiki-rendering-macro-box/src/test/resources/macrobox1.test
surli: Updated in another commit but forgot to push :)

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/MetaData.java
```diff
@@ -57,6 +57,13 @@
      */
     public static final String BASE = "base";
 
+    /**
+     * Represents an inline editable content.
+     *
+     * @since 10.9RC1
+     */
+    public static final String INLINE_EDITABLE = "wysiwyg-editable-content";
```
tmortagne: Since you change the class name you should also change this into something less WYSIWYG specific.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/UnchangedContentBlock.java
```diff
@@ -0,0 +1,70 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * This block represents a part of a macro content which is not transformed during the macro execution.
+ *
+ * @version $Id$
+ * @since 10.9RC1
+ */
+@Unstable
+public class UnchangedContentBlock extends MetaDataBlock
```
tmortagne: Wasn't the plan to also indicate if the content come from the macro content or some macro parameter ?

surli: Yes for me it was again another step.

tmortagne: OK.

xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/internal/macro/box/DefaultBoxMacro.java
```diff
@@ -59,15 +61,20 @@
...
+        super("Box", DESCRIPTION, new DefaultContentDescriptor(CONTENT_DESCRIPTION, true,
+            Block.LIST_BLOCK_TYPE), BoxMacroParameters.class);
         setDefaultCategory(DEFAULT_CATEGORY_FORMATTING);
     }
 
     @Override
     protected List<Block> parseContent(P parameters, String content, MacroTransformationContext context)
         throws MacroExecutionException
     {
+
```
tmortagne: Should cleanup those empty lines.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/DefaultContentDescriptor.java
```diff
@@ -65,9 +74,22 @@ public DefaultContentDescriptor(boolean mandatory)
      * @param mandatory indicate if the content is mandatory.
      */
     public DefaultContentDescriptor(String description, boolean mandatory)
+    {
+        this(description, mandatory, null);
```
tmortagne: This should be consistent with the default implementation in `ContentDescriptor`. Either `null` (if we think it's better to make explicit the content type is unknown) for both or `String.class` (if we feel String.class to be more consistent with parameters types logic) for both.

surli:
@mflorea @tmortagne please have a special look on the diff of the test for macro Figures and Message: I'm not completely sure it's the expected result we want.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/test/resources/macromessage3.test
```diff
@@ -63,6 +63,8 @@ onSpace
 onWord [possibility]
 endListItem
 endList [BULLETED]
+beginMetaData [[unchanged-content]=[true]]
```
tmortagne: This does not make much sense to me. It should be around the content, not after it.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/test/resources/macromessage2.test
```diff
@@ -63,6 +63,8 @@ onSpace
 onWord [possibility]
 endListItem
 endList [BULLETED]
+beginMetaData [[unchanged-content]=[true]]
```
tmortagne: This does not make much sense to me. It should be around the content, not after it.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/test/resources/macromessage1.test
```diff
@@ -63,6 +63,8 @@ onSpace
 onWord [possibility]
 endListItem
 endList [BULLETED]
+beginMetaData [[unchanged-content]=[true]]
```
tmortagne: This does not make much sense to me. It should be around the content, not after it.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/test/resources/macromessage3.test
surli: I agree, it was one of my concern: this is because when doing a message, the content is put from the title parameter, not from its content. So basically it's a box with a title but with an empty content. See: https://github.com/xwiki/xwiki-rendering/blob/225ab59614552a72a5ebf48e332cc0a5ed352abc/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java#L80-L85

So I wonder how to fix that one properly

tmortagne: All I can think of is stop hacking the BoxMacro in AbstractMessageMacro the way we currently do (at least for inline content). It will be even worst when we start generating metadata for the title.

I guess one thing you could do quickly is to check in ContentMacro if the context macro really is a content macro, if it's not then you can't really trust the `content` execute gave you.

surli: > stop hacking the BoxMacro in AbstractMessageMacro the way we currently do (at least for inline content)

so you suggest I fix the MessageMacro to properly use the content of the box? 

> you could do quickly is to check in ContentMacro if the context macro really is a content macro

Just checked while executing ContentMacro tests and apparently it is.

tmortagne: > so you suggest I fix the MessageMacro to properly use the content of the box?

I suggest to stop using the ContentMacro for inline content or at least not that way (passing the content as title and an empty content) since it's a "lie".

> Just checked while executing ContentMacro tests and apparently it is.

I don't see how since AbstractMessageMacro does not modify the macro context. I'm referring to the following logic: `context.getCurrentMacroBlock().getId().equals("content")`

surli: Sorry but I don't follow there: 

> I suggest to stop using the ContentMacro for inline content

I don't see where ContentMacro is used here. It's BoxMacro that is used.

tmortagne: Yes box macro sorry, just a typo.

surli: ok but maybe this PR is not the better place to do that refactoring properly. Especially since the whole behaviour of MacroMessage is currently relying on Box Macro. 
Do we open a ticket about that?

tmortagne: > ok but maybe this PR is not the better place to do that refactoring properly. Especially since the whole behaviour of MacroMessage is currently relying on Box Macro.

That's why I suggested you to add a protection in the macro used by AbstractMessageMacro (which I called the  ContentMacro instead of DefaultBoxMacro) to not end up with meaningless metadata.

surli: I opened https://jira.xwiki.org/browse/XRENDERING-524 and https://github.com/xwiki/xwiki-rendering/pull/148 to start working more properly on this issue.
I'll see what I can do for the protection you suggest.

surli: I added the protection, but then I had to decrease a bit the coverage (from 96 to 95%): this is because I now have two cases (if I'm in the context of box macro or if I'm not), and both cases are already tested, but the latter is tested inside the macro message now so not counted for jacoco locally in macro box.

tmortagne: In box macro module you could add a TestMacro which use the box macro execute directly and use that test macro in a test. That way you would make sure it works but any macro (not just message macro which is going to change anyway).

tmortagne:
On figure macro I'm not much of an expert. You probably know more than me on it by now...

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
```diff
@@ -86,12 +88,14 @@ public boolean supportsInlineMode()
         List<Block> result = Collections.emptyList();
 
         // If we're not inside a FigureBlock then don't do anything.
-        Block parent = context.getCurrentMacroBlock().getParent();
+        Block parent = context.getCurrentMacroBlock().getParent().getParent();
```
vmassol: Looks fragile and a bit weird, no? Also it's not the parent but the grandparent (cf variable name) ;)

vmassol: It requires explanations IMO (or better a new private method to find if we're inside a FigureBlock, i.e. traverse up till MacroMarkerBlock or something like this)

vmassol: BTW not your fault, it's my code, and it should have been better coded :) This is just revealing the fragility

surli: > Looks fragile and a bit weird, no? 

Completely. I wasn't even sure we should use an UnchangedContent block on this macro.
So maybe, as for MessageMacro, what we should do is to remove the UnchangedContent on this one and refactor it in another PR before trying to apply the new metadata. WDYT?

surli:
@mflorea do you think this one needs more work, or can I start implement its usage in WikiMacro on xwiki-platform?

mflorea:
@surli how is the ``UnchangedContentBlock`` rendered in HTML / Annotated XHTML?

tmortagne:
Metadata are not generating any HTML right now. @surli did you created a jira issue about that ?

surli:
> Metadata are not generating any HTML right now. @surli did you created a jira issue about that ?

No I did not.

tmortagne:
Given XRENDERING-518's title it could be part of it but it's starting to be big so better have another jira issue on which XRENDERING-518 depends.

surli:
Opened: https://jira.xwiki.org/browse/XRENDERING-527

Now I got an issue with this one, the question of @mflorea reminds me to check xwiki-rendering-standalone and I got the following error on it:

```
Caused by: java.lang.NoSuchFieldError: LIST_BLOCK_TYPE
	at org.xwiki.rendering.internal.macro.message.AbstractMessageMacro.&lt;init&gt;(AbstractMessageMacro.java:66)
	at org.xwiki.rendering.internal.macro.message.ErrorMessageMacro.&lt;init&gt;(ErrorMessageMacro.java:43)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
```

It looks like there is an issue somewhere caused by LIST_BLOCK_TYPE that I introduced. But I don't understand the error message here...

tmortagne:
It's probably because you forgot to build all your modification before building xwiki-rendering-standalone and it's using your version of message macro but the master version of rendering-api (which does not have the new LIST_BLOCK_TYPE field).

tmortagne:
What you are missing is probably https://github.com/xwiki/xwiki-rendering/tree/master/xwiki-rendering-legacy/xwiki-rendering-legacy-api (same system than oldcore and legacy-oldcore).

surli:
Indeed @tmortagne this was the problem. So everything's passing for this one.

surli:
@mflorea so all tests are running with this one (rebased on master), so I think it's ready.

mflorea:
Can you give an example of HTML produced by a macro that supports in-line content editing? Basically I want to know what is the input that the CKEditor gets. Let's take for instance:

    before
    
    {{info}}
    text
    {{/info}}

    after

What is the HTML produced by this? We also need to handle the save of the macro content when it is edited in-line, probably in a separate PR. @tmortagne suggested to modify the HTML parser to detect meta data blocks inside macro blocks and convert them to the wiki syntax specified in the meta data.

tmortagne:
> probably in a separate PR

I would make sense in this PR but it's already big so yes could be another PR after this one.

surli:
@mflorea here's what I got when running your example:

```
<p>before</p><!--startmacro:info|-||-|text--><div data-xwiki-unchanged-content="true" class="xwiki-metadata-container"><div class="box infomessage"><p>text</p></div></div><!--stopmacro--><p>after</p>
```

mflorea:
@surli thanks. So the important part is:

    
    <div data-xwiki-unchanged-content="true" class="xwiki-metadata-container">
      ...
    </div>
    

I think we're missing:

* The **content type** (e.g. ``java.util.List<org.xwiki.rendering.block.Block>``), which can be used to limit the type of content the user can insert. I know this information is available in the macro content descriptor, but the content descriptor is not available on the client side when the editor is loading. Making an AJAX request to get the content descriptor at that point is a pain because I would have to "pause" the loading until I get the response. Best for me is to have this information in the meta data block.
* The binding to the macro parameter name / content. I know that by default we consider the "unchanged content" to be bound to the macro content, but it would be nice to decide on a meta data that will specify the macro parameter name in the future.

surli:
> The content type (e.g. java.util.List<org.xwiki.rendering.block.Block>), 

and what would be the value of this metadata? Literally the name of the class to use, such as `java.util.List<org.xwiki.rendering.block.Block>`?

And if I sum up: for a macro to be editable inline, we have to:
  1. Specify its type in content descriptor as List<Block>
  2. Create a new metadatablock for the whole macro specifying the type of the content descriptor
  3? Specify the syntax of the macro in the same metadatablock
  4. Create (maybe another) metadatablock to specify part of the content which does not change (and thus is editable)
  
I really think that step 2 should be done automatically for all the macro, based on their content descriptor. WDYT? 

> it would be nice to decide on a meta data that will specify the macro parameter name in the future

Why not using:
  * data-xwiki-unchanged-content: for the content
  * data-xwiki-unchanged-<parameter_name>: for a parameter

WDYT?

mflorea:
> and what would be the value of this metadata?

Exactly the value returned by ``ContentDescriptor#getType()``. And if the meta data has to be a string then you can call ``toString()`` on the returned ``Type``.

> Create a new metadatablock for the whole macro specifying the type of the content descriptor

Why a new meta data block? The idea is to have the content type where the content is. And since we already have a meta data block that wraps the content and specifies (among other things) the content syntax, it could also specify the content type. And when we add support for editing macro parameters in-line each meta data bock that corresponds to such a parameter will specify the parameter syntax and content type.

> data-xwiki-unchanged-<parameter_name>: for a parameter

I'm not sure, but maybe the macro parameter name supports characters that are not allowed in XML attribute names. I would rather have the parameter name as the attribute value.

surli:
> And since we already have a meta data block that wraps the content and specifies (among other things) the content syntax,

That's currently true only if we use the content macro. Do you think that's enough? 

> I'm not sure, but maybe the macro parameter name supports characters that are not allowed in XML attribute names. I would rather have the parameter name as the attribute value.

Yeah I thought about this after posting my comment. So we can use `data-xwiki-parameter-name="<name of parameter>"` to get the parameter name and reuse `data-xwiki-unchanged-content`: if the parameter name is not given, then it's a metadata for the content itself, else it's for the parameter's content. WDYT?

mflorea:
> That's currently true only if we use the content macro. Do you think that's enough?

Is the Info Macro using the Content Macro? Because in the example above you indicated that the Info Macro generates the ``xwiki-metadata-container``. So to be clear, what I would like to get is:

    
    <div data-xwiki-unchanged-content="true"
        data-xwiki-content-type="java.util.List<org.xwiki.rendering.block.Block>"
        class="xwiki-metadata-container">
      ...
    </div>
    

or even

    <div data-xwiki-unchanged-content="java.util.List<org.xwiki.rendering.block.Block>"
        class="xwiki-metadata-container">

i.e. the ``unchanged-content`` meta data could indicate the content type.

surli:
> Is the Info Macro using the Content Macro?

No but here I was talking about the syntax of the content: as you can see it's not available in the metadata of snippet I gave to you.
For the rest of your answer, it's not a problem.

tmortagne:
> reuse data-xwiki-unchanged-content

If the it's reused both both content and parameters `data-xwiki-unchanged` might be less misleading.

tmortagne:
> Exactly the value returned by ContentDescriptor#getType(). And if the meta data has to be a string then you can call toString() on the returned Type

Yes toString() works fine. If you want to be safe you can go trough ConverterManager component which will serialize it in a way that can be parsed back to Type (right now it means toString() but in case toString() implementation change since it not an API...).

tmortagne:
> <div data-xwiki-unchanged-content="java.util.List<org.xwiki.rendering.block.Block>"
    class="xwiki-metadata-container">

Sounds good.

surli:
@mflorea so I applied the changes you asked and refactor a bit my previous work. So the result we get now is:
```
<p>before</p>
<!--startmacro:info|-||-|text-->
<div data-xwiki-unchanged-content="java.util.List&lt; org.xwiki.rendering.block.Block &gt;" class="xwiki-metadata-container">
<div class="box infomessage"><p>text</p></div>
</div>
<!--stopmacro-->
<p>after</p>
```

mflorea:
I'm surprised by the whitespace around ``org.xwiki.rendering.block.Block`` but if the ``ConverterManager`` generates that then I'll have to use it on my side. Looks good otherwise. I'll let @tmortagne comment on the actual changes. I'll update my code in CKEditor to handle the HTML code produced by the rendering.

tmortagne:
As I said the ConverterManager just call toString() right now so I guess that's what Java see as the canonical form of the type.

surli:
@tmortagne could you please have a look on my last commit on this PR? It's the first implem to manage the macro content in the xhtml parser. All tests are passing, but I did not test yet using macro inside macro or macro inline.
I'm not really fond of the change made in `TagStack` but I didn't find another way to manage it.

tmortagne:
Would be ice to have a test with a macro not explicitly indicating the syntax (that's the case of 90% of the editable macros). For example an info macro inside a content macro.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/MetaData.java
```diff
@@ -140,4 +147,9 @@ public int hashCode()
     {
         return this.metadata.hashCode();
     }
+
+    @Override public String toString()
```
tmortagne: Not following XWiki codestyle.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XHTMLXWikiGeneratorListener.java
```diff
@@ -50,14 +50,14 @@
     /**
      * Defines the class to use when an element represents a MetaData block.
      *
-     * @since 10.9RC1
+     * @since 10.9
```
tmortagne: Wrong since.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiSpanTagHandler.java
```diff
@@ -32,51 +35,75 @@
...
+
+    private XWikiMacroHandler xWikiMacroHandler;
+
+    /**
+     * Default constructor of a {@link XWikiSpanTagHandler}.
+     *
+     * @param componentManager is used to retrieved the proper parser component for serializing an unchanged content
+     * @param parser the current parser is actually used to simplify the build of other parsers.
+     */
+    public XWikiSpanTagHandler(ComponentManager componentManager, XHTMLParser parser)
```
tmortagne: Would be better to add a `@since`.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
```diff
@@ -62,4 +62,19 @@
      * Stack parameter which hold the instance of the WikiModel listener bridge for link being parsed.
      */
     String LINK_LISTENER = "linkListener";
+
+    /**
+     * Stack parameter which hold the instances of the MacroInfo encountered.
+     */
+    String MACRO_INFO = "macroInfo";
```
tmortagne: Would be better to add a `@since`.

xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/xhtml/impl/TagStack.java
```diff
@@ -63,7 +70,12 @@
 
     private final Deque<Map<String, Object>> fStackParameters = new ArrayDeque<Map<String, Object>>();
 
-    private boolean fIgnoreElements;
+    /**
+     * A stack of rules to ignore elements
+     *
+     * @since 10.10RC1
```
tmortagne: Since tag does not really bring much for a private field.

tmortagne:
Only a few minor stuff left, cannot see anything else on my side.

surli:
I'll fix the minor stuff reported by @tmortagne but could you have a look too @mflorea to check if I'm not missing something here?

surli:
ping @mflorea

mflorea:
I don't know the rendering code that well so it's hard for me to review that part. As long as the output is what we discussed above then it should be OK for me on the CKEditor side. We'll see how it works end to end.

surli:
ok then just fixed @tmortagne reported issues. It's ready to be merged.

tmortagne:
Done.

surli:
Finally \o/ Thanks both of you for the help on this one!

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/InlineEditableContentBlock.java
vmassol: The fact that some block is not modified during execution doesn't mean that we want to make this content editable inline. For example you can have some static text that you don't want to be editable inline.

PR #146, opened 2018-10-01
XRENDERING-523: Set the syntax in Macro Content

surli:
## Issue

https://jira.xwiki.org/browse/XRENDERING-523

## Changes

  * Set the syntax directly in the parent node as parameter, if there's
only one
  * Add a new metadatablock with the syntax if there's multiples blocks
  * Add a new test to check this latter case

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-content/pom.xml
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-content/pom.xml
@@ -36,4 +36,13 @@
     <!-- Name to display by the Extension Manager -->
     <xwiki.extension.name>Content Macro</xwiki.extension.name>
   </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-syntax-html</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
 </project>
--- a/xwiki-rendering-macros/xwiki-rendering-macro-content/src/main/java/org/xwiki/rendering/internal/macro/content/ContentMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-content/src/main/java/org/xwiki/rendering/internal/macro/content/ContentMacro.java
@@ -91,18 +91,10 @@ public class ContentMacro extends AbstractMacro<ContentMacroParameters>
     {
         try {
             List<Block> blocks = getSyntaxParser(parameters.getSyntax()).parse(new StringReader(content)).getChildren();
+            MetaDataBlock metaDataBlock = new MetaDataBlock(blocks, MetaData.SYNTAX,
+                parameters.getSyntax().toIdString());
 
-            // if the result is only one metadata block, we put the syntax metadata in it
-            if (blocks.size() == 1 && blocks.get(0) instanceof MetaDataBlock) {
-                ((MetaDataBlock) blocks.get(0)).getMetaData()
-                    .addMetaData(MetaData.SYNTAX, parameters.getSyntax().toIdString());
-                return blocks;
-            // else we create a dedicated MetadataBlock
-            } else {
-                MetaDataBlock metaDataBlock = new MetaDataBlock(blocks, MetaData.SYNTAX,
-                    parameters.getSyntax().toIdString());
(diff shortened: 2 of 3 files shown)


surli:
Ready to review @tmortagne

xwiki-rendering-macros/xwiki-rendering-macro-content/src/main/java/org/xwiki/rendering/internal/macro/content/ContentMacro.java
```diff
@@ -87,7 +90,18 @@ public boolean supportsInlineMode()
         throws MacroExecutionException
     {
         try {
-            return getSyntaxParser(parameters.getSyntax()).parse(new StringReader(content)).getChildren();
+            List<Block> blocks = getSyntaxParser(parameters.getSyntax()).parse(new StringReader(content)).getChildren();
+
+            // if the result is only one block, we put the syntax metadata in it
+            if (blocks.size() == 1) {
+                blocks.get(0).setParameter(MetaData.SYNTAX, parameters.getSyntax().toIdString());
```
tmortagne: You are mixing two different things. Block parameters have nothing to do with MetaData, you should check if the Block is a MetaDataBlock.

surli: OK actually as the Javadoc of `Block#getParameter` was stating: 
> A Parameter is a generic key/value which can be used to add metadata to a block.

I assumed that it could avoid us to introduce a new MetadataBlock. But now I imagine it's not the same kind of metadata: parameters are considered only for "relevant" parameter of a block, such as the target for a link or an image. Is that right? (Sorry, still trying to remove some ambiguities I might have)

tmortagne: Yes we should be more careful in the javadoc and try to avoid using "metadata" words for things not related to the actual Metadata class even if it's correct in English.

surli: Thanks for the explanation. Edited the PR.

surli:
what about the new changes @tmortagne ?

xwiki-rendering-macros/xwiki-rendering-macro-content/src/main/java/org/xwiki/rendering/internal/macro/content/ContentMacro.java
```diff
@@ -87,7 +90,19 @@ public boolean supportsInlineMode()
         throws MacroExecutionException
     {
         try {
-            return getSyntaxParser(parameters.getSyntax()).parse(new StringReader(content)).getChildren();
+            List<Block> blocks = getSyntaxParser(parameters.getSyntax()).parse(new StringReader(content)).getChildren();
+
+            // if the result is only one metadata block, we put the syntax metadata in it
+            if (blocks.size() == 1 && blocks.get(0) instanceof MetaDataBlock) {
```
tmortagne: In practice there is almost no chance of this ever happening (you won't have that with all the parsers I know at least and I can't think of a reason for a parser to do that) so not sure it really worth it.

surli: Wasn't sure either and it's hard to test properly. I'll remove it.

surli:
@tmortagne since its BFD, maybe a good day to merge this one? ;)

tmortagne:
Indeed sorry I totally forgot about this pull request.

xwiki-rendering-macros/xwiki-rendering-macro-content/pom.xml
```diff
@@ -36,4 +36,13 @@
...
   </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-syntax-html</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
```
vmassol: you shouldn't need this AFAICS.

PR #148, opened 2018-10-02
XRENDERING-524: Message Macro should not put all its content in the title of macro box

surli:


surli:
@vmassol apparently it was your code so maybe you know why you were relying on this behaviour?

PR #150, opened 2018-10-08
XRENDERING-527: HTML renderer should handle metadata to create dedicated attribute

surli:
* Create a new dedicated block for metadata in
AnnotatedXHTMLChainingRenderer

### DIFF
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
@@ -19,9 +19,10 @@
  */
 package org.xwiki.rendering.internal.renderer.xhtml;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
+import org.xwiki.rendering.internal.parser.xhtml.wikimodel.XHTMLXWikiGeneratorListener;
 import org.xwiki.rendering.internal.renderer.xhtml.image.XHTMLImageRenderer;
 import org.xwiki.rendering.internal.renderer.xhtml.link.XHTMLLinkRenderer;
 import org.xwiki.rendering.listener.MetaData;
@@ -36,8 +37,6 @@ import org.xwiki.rendering.listener.chaining.ListenerChain;
  */
 public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
 {
-    private static final String METADATA_BLOCK_ELEMENT = "span";
-
     /**
      * Renders a Macro definition into Annotated XHTML.
      */
@@ -88,24 +87,36 @@ public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
         }
     }
 
+    /**
+     * @return a span element if we are inside an inline macro. Else a div.
+     */
+    private String getMetadataContainerElement()
+    {
+        if (getBlockState().isInLine()) {
+            return "span";
+        } else {
+            return "div";
+        }
+    }
+
     @Override
     public void beginMetaData(MetaData metadata)
     {
-
-        Map<String, String> parameters = new HashMap<>();
+        Map<String, String> attributes = new LinkedHashMap<>();
 
         for (Map.Entry<String, Object> metadaPair : metadata.getMetaData().entrySet()) {
-            parameters.put(metadaPair.getKey(), metadaPair.getValue().toString());
+            attributes.put(XHTMLXWikiGeneratorListener.METADATA_ATTRIBUTE_PREFIX + metadaPair.getKey(),
+                metadaPair.getValue().toString());
         }
 
-        parameters.put("class", "xwiki-metadata-block");
+        attributes.put("class", XHTMLXWikiGeneratorListener.METADATA_CONTAINER_CLASS);
 
-        this.getXHTMLWikiPrinter().printXMLStartElement(METADATA_BLOCK_ELEMENT, parameters);
+        this.getXHTMLWikiPrinter().printXMLStartElement(getMetadataContainerElement(), attributes);
     }
 
     @Override
     public void endMetaData(MetaData metadata)
     {
-        getXHTMLWikiPrinter().printXMLEndElement(METADATA_BLOCK_ELEMENT);
+        getXHTMLWikiPrinter().printXMLEndElement(getMetadataContainerElement());
     }
 }
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
@@ -104,9 +104,9 @@ public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
     {
         Map<String, String> attributes = new LinkedHashMap<>();
 
-        for (Map.Entry<String, Object> metadaPair : metadata.getMetaData().entrySet()) {
-            attributes.put(XHTMLXWikiGeneratorListener.METADATA_ATTRIBUTE_PREFIX + metadaPair.getKey(),
-                metadaPair.getValue().toString());
+        for (Map.Entry<String, Object> metadataPair : metadata.getMetaData().entrySet()) {
+            attributes.put(XHTMLXWikiGeneratorListener.METADATA_ATTRIBUTE_PREFIX + metadataPair.getKey(),
+                metadataPair.getValue().toString());
         }
 
         attributes.put("class", XHTMLXWikiGeneratorListener.METADATA_CONTAINER_CLASS);
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
@@ -19,10 +19,12 @@
  */
 package org.xwiki.rendering.internal.renderer.xhtml;
 
+import java.util.HashMap;
 import java.util.Map;
 
 import org.xwiki.rendering.internal.renderer.xhtml.image.XHTMLImageRenderer;
 import org.xwiki.rendering.internal.renderer.xhtml.link.XHTMLLinkRenderer;
+import org.xwiki.rendering.listener.MetaData;
 import org.xwiki.rendering.listener.chaining.ListenerChain;
 
 /**
@@ -34,6 +36,8 @@ import org.xwiki.rendering.listener.chaining.ListenerChain;
  */
 public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
 {
+    private static final String METADATA_BLOCK_ELEMENT = "span";
+
     /**
      * Renders a Macro definition into Annotated XHTML.
      */
@@ -82,5 +86,26 @@ public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
             // so that the macro can be reconstructed when moving back from XHTML to XDOM.
             this.macroRenderer.endRender(getXHTMLWikiPrinter());
         }
+    }
+
+    @Override
+    public void beginMetaData(MetaData metadata)
+    {
+
+        Map<String, String> parameters = new HashMap<>();
+
+        for (Map.Entry<String, Object> metadaPair : metadata.getMetaData().entrySet()) {
+            parameters.put(metadaPair.getKey(), metadaPair.getValue().toString());
+        }
+
+        parameters.put("class", "xwiki-metadata-block");
+
+        this.getXHTMLWikiPrinter().printXMLStartElement(METADATA_BLOCK_ELEMENT, parameters);
+    }
+
+    @Override
+    public void endMetaData(MetaData metadata)
+    {
+        getXHTMLWikiPrinter().printXMLEndElement(METADATA_BLOCK_ELEMENT);
     }
 }
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XHTMLXWikiGeneratorListener.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XHTMLXWikiGeneratorListener.java
@@ -26,6 +26,7 @@ import java.util.regex.Pattern;
 import org.apache.commons.lang3.tuple.Pair;
 import org.xwiki.rendering.internal.parser.wikimodel.DefaultXWikiGeneratorListener;
 import org.xwiki.rendering.listener.Listener;
+import org.xwiki.rendering.listener.MetaData;
 import org.xwiki.rendering.listener.reference.ResourceReference;
 import org.xwiki.rendering.listener.reference.ResourceType;
 import org.xwiki.rendering.parser.ResourceReferenceParser;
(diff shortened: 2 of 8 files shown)


surli:
@mflorea @tmortagne this one is a WIP. I'm actually not sure that's what you had in mind.
I thought it would be better to put the metadata in an existing block, but it seems to me actually quite complicated to do without rewriting entirely the renderers. That's why I decided to put a new block with the metadata, as it's actually represented in the macro. 

Now I'm not sure how to test it properly: the current tests in the renderer are using a `CompatibilityTestSuite` and it looks like it uses both a parser and a renderer to process the test. So I assume I would need to put some metadata in the examples `xwiki-rendering-test` and to also manage the parsing of metadata?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -34,6 +36,8 @@
  */
 public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
 {
+    private static final String METADATA_BLOCK_ELEMENT = "span";
```
mflorea: The element used to hold the meta data must depend on the wrapped / inner / nested content. ``span`` is good only if the wrapped content is in-line, otherwise, for block-level content, you must use ``div``. Imagine that the macro outputs a ``table``, it wouldn't be valid to have the ``table`` inside a ``span``.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -83,4 +87,25 @@ public void endMacroMarker(String name, Map<String, String> parameters, String c
             this.macroRenderer.endRender(getXHTMLWikiPrinter());
         }
     }
+
+    @Override
+    public void beginMetaData(MetaData metadata)
+    {
+
+        Map<String, String> parameters = new HashMap<>();
```
mflorea: I would call them ``attributes`` because that's how they are referred to in the XML / HTML world.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -83,4 +87,25 @@ public void endMacroMarker(String name, Map<String, String> parameters, String c
...
+
+        for (Map.Entry<String, Object> metadaPair : metadata.getMetaData().entrySet()) {
+            parameters.put(metadaPair.getKey(), metadaPair.getValue().toString());
```
mflorea: Shouldn't you generate ``data-`` attributes? And, if you don't use any namespace then how will you know what is meta data and what is simply HTML attributes added by some other code, when you parse the meta data back into XDOM?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
surli: OK it's actually what I thought: then I have to manage a context in the AnnotatedXHTMLChainingRenderer to flag if I'm between begin/endMacroMarker and what is the inline value, right?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
surli: Indeed

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
surli: > Shouldn't you generate data- attributes?

I thought there was already a processing for non-standard HTML attributes in the printer, so yeah I should :)

> And, if you don't use any namespace then how will you know what is meta data and what is simply HTML attributes added by some other code, when you parse the meta data back into XDOM?

Here, by namespace, you mean a dedicated name, right? Like `data-xwiki-...`, or you mean something more sophisticated such as a DTD? Because on the second case I'm not sure how to implement it.

mflorea: > Here, by namespace, you mean a dedicated name, right? Like ``data-xwiki-``

Yes, maybe not exactly ``data-xwiki-*`` but in any case something to distinguish the attributes that are mapped to XDOM meta data from attributes set by some other code (you never know). Otherwise you end up with all the ``data-`` attributes in the XDOM meta data and this may cause unexpected behavior.

Otherwise, you could write / read only known meta data (instead of everything), but then we would need to update the code when we add a new meta data (which doesn't happen very often, but still).

surli: Why not `data-xwiki-*` then? 

> Otherwise, you could write / read only known meta data (instead of everything), but then we would need to update the code when we add a new meta data (which doesn't happen very often, but still).

I could use some reflexivity here: the know metadata should be declared as public String in MetaData.class. WDYT?

surli: Hm actually maybe not a good idea, we don't use prefixed attribute to specify that it's a medata so a public string in the class could be anything else.

surli: Actually what you imply is that someone could write something like:
```
<div data-syntax="xwiki/2.1" data-custom="foo">
[HTML]
</div>
```
and we should be able to distinguish our own metadata `data-syntax` and a custom one.
But how would we parse such an example? 

```
BeginGroupBlock [[custom]=[foo]]
BeginMetaDataBlock [[syntax]=[xwiki/2.1]]
[PARSED HTML]
EndMetaDataBlock [[syntax]=[xwiki/2.1]]
EndGroupBlock [[custom]=[foo]]
```

but then when rendering this again we would obtain:
```
<div data-custom="foo">
<div data-syntax="xwiki/2.1">
[HTML]
</div>
</div>
```

So I'd say on the contrary that the div or span we create for the metadata use a dedicated class or id which specify that it's an element created to support XWiki metadata only. Any other custom attribute should be placed in another element. WDYT?

mflorea: My suggestion was to ignore the custom attributes. For instance, if you parse

    <div class="wikimodel-emptyline" data-custom="foo">

the custom attribute is lost AFAIK.

surli: OK I misunderstood but now we're on the same page :)

surli:
@mflorea so I implemented the parsing, some tests are still missing though.

surli:
OK now tests have been added and everything's running. So it's ready for review @mflorea @tmortagne.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/MetaData.java
```diff
@@ -36,6 +36,16 @@
      */
     public static final MetaData EMPTY = new MetaData();
 
+    /**
+     * Defines the class to use when an element represents a MetaData block.
+     */
+    public static final String METADATA_CONTAINER_CLASS = "xwiki-metadata-container";
```
tmortagne: This is a HTML thing, nothing to do here.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/MetaData.java
```diff
@@ -36,6 +36,16 @@
...
+
+    /**
+     * Defines the prefix to use for attribute which contain a metadata.
+     */
+    public static final String METADATA_ATTRIBUTE_PREFIX = "data-xwiki-";
```
tmortagne: Same comment.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -34,6 +36,11 @@
  */
 public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
 {
+    /**
+     * Returns true if inside an inline macro.
+     */
+    private boolean isInlineMacro;
```
tmortagne: No need. See XHTMLChainingRenderer#onId.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -82,5 +90,40 @@ public void endMacroMarker(String name, Map<String, String> parameters, String c
...
+            return "span";
+        } else {
+            return "div";
+        }
+    }
+
+    @Override
+    public void beginMetaData(MetaData metadata)
+    {
+
```
tmortagne: You should remove that empty line.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -82,5 +90,40 @@ public void endMacroMarker(String name, Map<String, String> parameters, String c
...
+        Map<String, String> attributes = new HashMap<>();
```
tmortagne: I would use a LinkedHashMap to not change the order.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -34,6 +36,11 @@
  */
 public class AnnotatedXHTMLChainingRenderer extends XHTMLChainingRenderer
```
tmortagne: There is also metadatas in beginDocument.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/MetaData.java
surli: I put it in there because I need the constant both in `xwiki-rendering-syntax-xhtml` and in `xwiki-rendering-syntax-annotatedxhtml` (for the renderer and the parser). Do you have any suggestion where to put else?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
surli: Not sure to understand the comment here. You want me to ignore the metadata processed by begin/endDocument? Or on the contrary to process them here?

tmortagne: Actually forget that. Handling begin/endDocument might do more harm than good right now, we'll see how it goes and if we need it later for some use case.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
```diff
@@ -83,4 +86,37 @@ public void endMacroMarker(String name, Map<String, String> parameters, String c
...
+            return "div";
+        }
+    }
+
+    @Override
+    public void beginMetaData(MetaData metadata)
+    {
+        Map<String, String> attributes = new LinkedHashMap<>();
+
+        for (Map.Entry<String, Object> metadaPair : metadata.getMetaData().entrySet()) {
```
mflorea: ``metada``?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XHTMLXWikiGeneratorListener.java
```diff
@@ -145,4 +165,70 @@ private ResourceReference computeResourceReference(String rawReference)
...
         return reference;
     }
+
+    private boolean isMetaDataElement(WikiParameters parameters)
+    {
+        if (parameters.getParameter(CLASS_ATTRIBUTE) != null) {
+            return METADATA_CONTAINER_CLASS.equals(parameters.getParameter(CLASS_ATTRIBUTE).getValue());
+        }
+
+        return false;
```
mflorea: This can be written in a single line:

    return parameters.getParameter(CLASS_ATTRIBUTE) != null &&
        METADATA_CONTAINER_CLASS.equals(parameters.getParameter(CLASS_ATTRIBUTE).getValue());

surli: Indeed, I'll simplify

xwiki-rendering-syntaxes/xwiki-rendering-syntax-annotatedxhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/AnnotatedXHTMLChainingRenderer.java
surli: :D I took 5 minutes to understand your question, apparently my brain reads directly `metadata` from `metada` now ^^

surli:
> Looks OK but I would like to know what the CKEditor gets as input for a macro with content editable in-line.

It would be: 
```
<span data-xwiki-syntax="xwiki/2.0" class="xwiki-metadata-container">inline content</span>
```

I did not manage to produce easily a test to output that.

tmortagne:
`data-xwiki-syntax="xwiki/2.0"` will only be there for macros not using the current syntax AFAIK.

mflorea:
The CKEditor needs more metadata:

* the content type
* the mapping to the macro content or parameter

surli:
Actually you have a test for an inline, see: 
https://github.com/xwiki/xwiki-rendering/pull/150/files#diff-cecfc52be1589324f7f8111849cd830d

surli:
> The CKEditor needs more metadata:

    the content type
    the mapping to the macro content or parameter

Here it's only a PR to manage metadata, it's not directly related to macro: I'll update https://github.com/xwiki/xwiki-rendering/pull/143 on macro after this one is merged.

mflorea:
OK, I'm just making sure we're not missing something.

surli:
> OK, I'm just making sure we're not missing something.

you're right :) 
Here the goal is only to display any kind of metadata in the HTML. I will use it for the macro in the side PR. 

I fixed the typo and the method so for me it's ok.

tmortagne:
Merged, thanks @surli !

PR #151, opened 2018-10-12
XRENDERING-527: HTML renderer should handle metadata to create dedicated attribute

surli:
Checked by running `mvn clean install -Pquality`.


surli:
@vmassol @tmortagne this should fix the build.

vmassol:
@surli Isn't there an issue that caused this?

PR #153, opened 2018-11-06

surli:
* Fix the mutation threshold
  * Add some test inside the module


vmassol:
If you added more tests maybe the jacoco threshold can be increased too?

PR #155, opened 2018-11-15

surli:
* Macro marker are always printed for nested macros in annotatedxhtml


surli:
All tests are passing @mflorea

PR #156, opened 2018-11-16

surli:
* if no syntax metadata is available the fallback syntax is retrieved
from the RenderingContext#targetSyntax. If this latter is not available either, the final fallback is the parser syntax.

### DIFF
--- a/xwiki-rendering-integration-tests/src/test/java/org/xwiki/rendering/SimpleIntegrationTests.java
+++ b/xwiki-rendering-integration-tests/src/test/java/org/xwiki/rendering/SimpleIntegrationTests.java
@@ -20,8 +20,14 @@
 package org.xwiki.rendering;
 
 import org.junit.runner.RunWith;
+import org.xwiki.context.ExecutionContext;
+import org.xwiki.context.ExecutionContextManager;
+import org.xwiki.rendering.internal.transformation.MutableRenderingContext;
+import org.xwiki.rendering.syntax.Syntax;
 import org.xwiki.rendering.test.integration.RenderingTestSuite;
+import org.xwiki.rendering.transformation.RenderingContext;
 import org.xwiki.test.annotation.AllComponents;
+import org.xwiki.test.mockito.MockitoComponentManager;
 
 /**
  * Run all tests found in {@code simple/*.test} files located in the classpath. These {@code *.test} files must follow
@@ -35,4 +41,17 @@ import org.xwiki.test.annotation.AllComponents;
 @AllComponents
 public class SimpleIntegrationTests
 {
+    @RenderingTestSuite.Initialized
+    public void initialize(MockitoComponentManager componentManager) throws Exception
+    {
+        ExecutionContext executionContext = new ExecutionContext();
+        ExecutionContextManager executionContextManager = componentManager.getInstance(ExecutionContextManager.class);
+        executionContextManager.initialize(executionContext);
+
+        // Set TargetSyntax for Macro tests
+        RenderingContext renderingContext = componentManager.getInstance(RenderingContext.class);
+        ((MutableRenderingContext) renderingContext).push(renderingContext.getTransformation(),
+            renderingContext.getXDOM(), renderingContext.getDefaultSyntax(), renderingContext.getTransformationId(),
+            renderingContext.isRestricted(), Syntax.XWIKI_2_0);
+    }
 }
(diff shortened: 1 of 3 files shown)


xwiki-rendering-integration-tests/src/test/java/org/xwiki/rendering/SimpleIntegrationTests.java
```diff
@@ -35,4 +41,17 @@
 @AllComponents
 public class SimpleIntegrationTests
 {
+    @RenderingTestSuite.Initialized
+    public void initialize(MockitoComponentManager componentManager) throws Exception
```
tmortagne: I really don't think this in the right place:
* this only cover a subset of the rendering tests
* you can't remove the ExecutionContext after the test execution this way

It would really make much more sense to add before/after code in both RenderingTest right were the code which convert from one syntax to another (and execute tx sometimes) is located.

xwiki-rendering-integration-tests/src/test/resources/simple/macros/macro20.test
```diff
@@ -37,7 +37,7 @@ onWord [paragraph]
 onSpecialSymbol [.]
 endParagraph
 onMacroStandalone [testcontentmacro] [] [{{testcontentmacro}}
-<p><strong>New content</strong></p>
+**New content**
```
tmortagne: So you did had a test after all, was just not expecting the right result :)

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiMacroHandler.java
```diff
@@ -62,6 +70,24 @@ public XWikiMacroHandler(ComponentManager componentManager, XHTMLParser parser)
         this.parser = parser;
     }
 
+    private String getSyntax(TagContext context) throws ComponentLookupException
+    {
+        if (context.getTagStack().getStackParameter(CURRENT_SYNTAX) != null) {
+            return (String) context.getTagStack().popStackParameter(CURRENT_SYNTAX);
+
+            // if the current syntax is not retrieved from the context, we get it from the RenderingContext
```
tmortagne: You might want to rewrite a little the first part since "contex" vs "RenderingContext" is a bit confusing. Maybe something like "previous nodes" instead.

xwiki-rendering-integration-tests/src/test/resources/simple/macros/macro20.test
surli: yep

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiMacroHandler.java
```diff
@@ -62,6 +70,24 @@ public XWikiMacroHandler(ComponentManager componentManager, XHTMLParser parser)
         this.parser = parser;
     }
 
+    private String getSyntax(TagContext context) throws ComponentLookupException
```
tmortagne: I think I would deal with the exception where getInstance is called instead and fallback on parser syntax.

xwiki-rendering-integration-tests/src/test/resources/simple/macros/macro20.test
surli: Actually it's wrong, the input is `.input|xhtml/1.0` so it should remain  as it was: here it changes because I forced the target syntax to xwiki/2.0

tmortagne: So inhering parent node was not working properly and you fixed another bug ?

surli: No it's just that the targetSyntax here should be xhtml, so the result should remain:
```
<p><strong>New content</strong></p>
```

Here the change is coming because I was forcing the targetSyntax as `xwiki/2.0` and I was not using what's defined in the test.
Now as I'm putting the RenderingContext elsewhere I'm using the targetSyntax using info from the test, so not forcing it anymore.

tmortagne: I see so that test was actually right and this change need to be reverted (and a new test added).

surli: I need to be reverted. Now I don't know how to write the test we want: how to express using annotatedxhtml that I'm in the context of a page using xwiki/2.0 in our rendering test framework?

tmortagne: It's simple a test which convert annotatedhtml (input) content into xwiki/2.1 (expect) content. In this use case the target syntax will be xwiki/2.1 with the change to are making to the rendering framework.

PR #157, opened 2018-11-16

surli:
* Also skip groups in InlineFilterListener


surli:
ping @tmortagne

PR #158, opened 2018-11-16

surli:
* Fix the macro message to put the unchanged content at the right
place

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/internal/macro/box/DefaultBoxMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/internal/macro/box/DefaultBoxMacro.java
@@ -49,6 +49,7 @@ public class DefaultBoxMacro<P extends BoxMacroParameters> extends AbstractBoxMa
 {
     /**
      * All the macro that needs to display an unchanged content div around the content.
+     * // TODO: this nasty hack will be removed as soon as message macro is refactored to use the AbstractBoxMacro
      */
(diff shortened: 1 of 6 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/internal/macro/box/DefaultBoxMacro.java
```diff
@@ -46,6 +47,12 @@
 @Singleton
 public class DefaultBoxMacro<P extends BoxMacroParameters> extends AbstractBoxMacro<P>
 {
+    /**
+     * All the macro that needs to display an unchanged content div around the content.
+     */
+    private static final List<String> ACCEPTED_MACRO_UNCHANGED_CONTENT = Arrays.asList("box", "info", "error",
+        "warning", "success");
```
vmassol: That's really bad and a no go for me. It means that the BoxMacro knows about the message macros which is really bad.

vmassol: Just to be clear: it's a no go unless we will fix it soon, i.e missing a todo comment to explain what should/will be done :)

surli: Just added the comments

tmortagne: We should definitely not have that in 10.10 final.

PR #159, opened 2018-11-19
XRENDERING-536: Message macro should rely on AbstractBoxMacro

surli:
* Message macro now relies on AbstractBoxMacro

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
@@ -85,6 +85,6 @@ public abstract class AbstractMessageMacro extends AbstractBoxMacro<BoxMacroPara
     @Override
     protected String getClassProperty()
     {
-        return super.getClassProperty() + ' ' + this.getDescriptor().getId() + "message";
+        return super.getClassProperty() + ' ' + this.getDescriptor().getId().getId() + "message";
     }
 }
--- a/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
@@ -39,11 +39,12 @@ import org.xwiki.rendering.transformation.MacroTransformationContext;
  */
 public abstract class AbstractMessageMacro extends AbstractBoxMacro<BoxMacroParameters>
 {
-
     /**
      * Predefined error message.
      */
     public static final String CONTENT_MISSING_ERROR = "The required content is missing.";
+
+    private MacroTransformationContext currentContext;
 
     /**
      * Create and initialize the descriptor of the macro.
@@ -65,8 +66,7 @@ public abstract class AbstractMessageMacro extends AbstractBoxMacro<BoxMacroPara
         if (StringUtils.isEmpty(content)) {
             throw new MacroExecutionException(CONTENT_MISSING_ERROR);
         }
-
-        parameters.setCssClass(context.getCurrentMacroBlock().getId() + "message");
+        this.currentContext = context;
         return super.execute(parameters, content, context);
     }
 
@@ -83,5 +83,11 @@ public abstract class AbstractMessageMacro extends AbstractBoxMacro<BoxMacroPara
     public boolean supportsInlineMode()
     {
         return true;
+    }
+
+    @Override
+    protected String getClassProperty()
+    {
+        return super.getClassProperty() + ' ' + this.currentContext.getCurrentMacroBlock().getId() + "message";
     }
 }
--- a/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
@@ -57,25 +57,12 @@ public abstract class AbstractMessageMacro extends AbstractBoxMacro<BoxMacroPara
     }
 
     @Override
-    public List<Block> execute(BoxMacroParameters parameters, String content,
-        MacroTransformationContext context) throws MacroExecutionException
-    {
-        return super.execute(parameters, content, context);
-    }
-
-    @Override
     protected List<Block> parseContent(BoxMacroParameters parameters, String content,
         MacroTransformationContext context) throws MacroExecutionException
     {
         List<Block> macroContent = getMacroContentParser().parse(content, context, false, context.isInline())
             .getChildren();
         return Collections.singletonList(new MetaDataBlock(macroContent, this.getUnchangedContentMetaData()));
-    }
-
-    @Override
-    public boolean supportsInlineMode()
-    {
-        return true;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


surli:
ping @vmassol @tmortagne you might want to review this one.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -62,23 +54,29 @@
...
+        MacroTransformationContext context) throws MacroExecutionException
     {
         if (StringUtils.isEmpty(content)) {
             throw new MacroExecutionException(CONTENT_MISSING_ERROR);
         }
 
-        BoxMacroParameters boxParameters = new BoxMacroParameters();
-
-        boxParameters.setCssClass(context.getCurrentMacroBlock().getId() + "message");
+        parameters.setCssClass(context.getCurrentMacroBlock().getId() + "message");
```
tmortagne: I'm not a big fan of modifying the input parameters object.

A cleaner solution might be to overwrite getClassProperty() and return something like `super.getClassProperty() + ' ' + context.getCurrentMacroBlock().getId() + "message"`. This also allow message macro used to indicate a custom class which is a nice news feature.

tmortagne: Not at all. I did not suggested to introduce anything new, just to not break BoxMacroParameters properties.

surli: Not sure to follow. Here the class is dedicated to message macro, as the name suggest, so it feels wrong to put it as default for box macro...

surli: My bad I mixed up between getClassProperty and getCssClass. OK I'll do that.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -62,28 +55,39 @@
...
     @Override
     public boolean supportsInlineMode()
     {
         return true;
     }
+
+    @Override
+    protected String getClassProperty()
+    {
+        return super.getClassProperty() + ' ' + this.currentContext.getCurrentMacroBlock().getId() + "message";
```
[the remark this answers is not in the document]
surli: actually it's based on the macro Id, but indeed it allows to get rid of the context.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -62,28 +53,38 @@
...
     @Override
     public boolean supportsInlineMode()
     {
         return true;
     }
+
+    @Override
+    protected String getClassProperty()
+    {
+        return super.getClassProperty() + ' ' + this.getDescriptor().getId() + "message";
```
tmortagne: The macro might be associated to a specific syntax to it's probably safer to do this.getDescriptor().getId().getId().

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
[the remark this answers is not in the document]
tmortagne: Yes getId() not getName() sorry.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -62,28 +53,38 @@
...
+            BoxMacroParameters.class);
     }
 
     @Override
-    public List<Block> execute(Object parameters, String content, MacroTransformationContext context)
-        throws MacroExecutionException
+    public List<Block> execute(BoxMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
     {
         if (StringUtils.isEmpty(content)) {
```
tmortagne: Might make more sense to move this check in parseContent, WDYT ?

surli: Actually parseContent is never called if the content is null, according to AbstractBoxMacro#execute. So if we want to keep the same behaviour as before, we might prefer to keep it there.

tmortagne: OK this is a weird behavior from AbstractBoxMacro, at the very least it take into account ContentDescriptor#isMandatory() and fail instead of just do nothing if true.

surli: so you suggest I change the behaviour of AbstractBoxMacro to check if the content is mandatory and to move the exception we were using in message directly in AbstractBoxMacro when the content is empty and mandatory?

tmortagne: I just checked DefaultBoxMacro and noticed that the content is defined as mandatory there too but never taken into account :)

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -62,28 +52,35 @@
...
         super(macroName, macroDescription,
-            new DefaultContentDescriptor("Content of the message", true, Block.LIST_BLOCK_TYPE));
+            new DefaultContentDescriptor("Content of the message", true, Block.LIST_BLOCK_TYPE),
+            BoxMacroParameters.class);
     }
 
     @Override
-    public List<Block> execute(Object parameters, String content, MacroTransformationContext context)
-        throws MacroExecutionException
+    public List<Block> execute(BoxMacroParameters parameters, String content,
```
tmortagne: So you don't need the whole method anything actually :)

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -62,28 +52,35 @@
...
+    protected List<Block> parseContent(BoxMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
+    {
+        List<Block> macroContent = getMacroContentParser().parse(content, context, false, context.isInline())
+            .getChildren();
+        return Collections.singletonList(new MetaDataBlock(macroContent, this.getUnchangedContentMetaData()));
     }
 
     @Override
     public boolean supportsInlineMode()
```
tmortagne: Actually this is the same in AbstractBoxMacro.

surli: In DefaultBoxMacro you mean.

surli: And we want to keep separation from DefaultBoxMacro.

tmortagne: > In DefaultBoxMacro you mean.

No I don't :)

PR #160, opened 2018-11-19
XRENDERING-538: Figure and FigureCaption macros should be editable inline

surli:
* Allow figure and FigureCaption to be editable inline. Change tests
accordingly.

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
@@ -30,6 +30,7 @@ import org.xwiki.component.annotation.Component;
 import org.xwiki.rendering.block.Block;
 import org.xwiki.rendering.block.FigureBlock;
 import org.xwiki.rendering.block.FigureCaptionBlock;
+import org.xwiki.rendering.block.MetaDataBlock;
 import org.xwiki.rendering.block.XDOM;
 import org.xwiki.rendering.macro.AbstractNoParameterMacro;
 import org.xwiki.rendering.macro.MacroContentParser;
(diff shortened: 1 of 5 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
```diff
@@ -86,13 +88,23 @@ public boolean supportsInlineMode()
     {
         List<Block> result = Collections.emptyList();
 
+        Block figureBlock = context.getCurrentMacroBlock().getFirstBlock(new BlockMatcher()
```
tmortagne: Note that there is a ClassBlockMatcher to which you could pass FigureBlock.

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
```diff
@@ -86,13 +87,16 @@ public boolean supportsInlineMode()
     {
         List<Block> result = Collections.emptyList();
 
-        // If we're not inside a FigureBlock then don't do anything.
-        Block parent = context.getCurrentMacroBlock().getParent();
-        if (parent instanceof FigureBlock) {
+        Block parentBlock = context.getCurrentMacroBlock().getParent();
+
+        // we don't want to do anything if we're not in a FigureBlock
```
vmassol: IMO this was not needed to be changed. And you put a lowercase for starting the sentence ;)

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
```diff
@@ -86,13 +87,16 @@ public boolean supportsInlineMode()
...
+        if (parentBlock instanceof FigureBlock) {
             XDOM xdom = this.contentParser.parse(content, context, false, false);
             List<Block> figureCaptionChildren = xdom.getChildren();
             this.parserUtils.removeTopLevelParagraph(figureCaptionChildren);
-            result = Collections.singletonList(new FigureCaptionBlock(figureCaptionChildren));
+            List<Block> contentBlock = Collections.singletonList(new FigureCaptionBlock(figureCaptionChildren));
+
+            result = Collections.singletonList(new MetaDataBlock(contentBlock, this.getUnchangedContentMetaData()));
```
vmassol: Small detail: there's no need to prefix with "this.". We do that only for fields normally.

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
tmortagne: Yes would probably be cleaner to go back to previous version since it does not looks like a obvious improvement.

vmassol:
Sounds good, except for the minor details I mentioned. thanks

PR #161, opened 2018-11-20

surli:
* Add missing unstable annotations


surli:
@gdelhumeau @tmortagne @vmassol

PR #165, opened 2018-11-23
XRENDERING-539: Change the "unchanged-content" metadata name

surli:
* Use "non-generated-content" as a metadata name
  * Fix the associated variables and methods names
  * Fix the comments
  * Change the order of metadata in macro-content test, since the
alphabetical order changed between syntax and non-generated-content.

### DIFF
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/AbstractMacro.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/AbstractMacro.java
@@ -257,15 +257,15 @@ public abstract class AbstractMacro<P> implements Macro<P>, Initializable
     }
 
     /**
-     * Helper to get the proper metadata for unchanged content (i.e. content that has not gone through a
-     * Transformation).
+     * Helper to get the proper metadata for non generated content (i.e. content that has not gone through a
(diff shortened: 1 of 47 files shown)


surli:
@tmortagne @vmassol

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/AbstractMacro.java
```diff
@@ -257,15 +257,15 @@ protected void setDefaultCategory(String defaultCategory)
...
     /**
-     * Helper to get the proper metadata for unchanged content (i.e. content that has not gone through a
-     * Transformation).
+     * Helper to get the proper metadata for non generated content (i.e. content that has not gone through a
+     * Transformation). This content can be used for inline editing.
      *
      * @return the new metadata with the content type for the content represented as a string (e.g.
-     *         {@code java.util.List< org.xwiki.rendering.block.Block >} for content of type {@code Listw<Block>}
+     *         {@code java.util.List< org.xwiki.rendering.block.Block >} for content of type {@code List<Block>}
      * @since 10.10RC1
```
tmortagne: Wrong since.

vmassol: why? the method signature has no changed AFAICS.

PR #167, opened 2018-11-27
XWIKI-15914: Cannot add Attachment Selector Macro into a page

surli:
* Protect XWikiCommentHandler#handleMacroCommentStop in case no macro
info is available in the context


tmortagne:
Would be better to add a test for it with the XHTML which trigger this crash.

surli:
> Would be better to add a test for it with the XHTML which trigger this crash.

Agreed but I'm not sure I can easily: 
  1. I think we got the error because of the problem in CKEditor
  2. the involved macro is not available in the xwiki-rendering

tmortagne:
I don't see why, just copy paste the XHTML which triggered the bug in the XHTML -> events test. You don't need the macro (you can change the name in the XHTML to make the test more generic) since it's not going to be executed.

surli:
Here's the test which trigger the error.

PR #168, opened 2018-12-04
XRENDERING-544: {{info}} macro with a single paragraph content generates a different XDOM than before

surli:
- Implement specific behaviour when message macro have a single
paragraph content


surli:
ready for review @vmassol

xwiki-rendering-macros/xwiki-rendering-macro-message/src/test/resources/macromessage5.test
```diff
@@ -0,0 +1,24 @@
+.runTransformations
+.#-----------------------------------------------------
+.input|xwiki/2.0
+.# Validate that the paragraph is stripped when it's not needed
+.#-----------------------------------------------------
+{{info}}Only **simple** message{{/info}}
```
vmassol: You don't need bold to validate this test. You just need a single word actually for a minimal test

vmassol: Yes please. My POV is that:
* a test should test what to says it's testing
* it should be as simple as possible since otherwise it raises question. In this case, if I'm reading this test, I'm asking myself why is there a bold element and why is that important for the test.

Do you agree?

surli: As I really want to test here the behaviour both for the simple paragraph and what happens with styling in it (not tested elsewhere), I propose a middle-way: I keep this test with a more complete comment, and I add a new simpler one, just for the paragraph.

vmassol: > As I really want to test here the behaviour both for the simple paragraph and what happens with styling in it (not tested elsewhere),

Why? We don't any code in the message macro that has any impact on that? Am I missing something?

surli: > We don't have any code in the message macro that has any impact on that? Am I missing something?

Nop, just me starting to be a bit paranoïd with macro message, and wanted to cover all cases :) 
OK I'll remove it.

vmassol: > Nop, just me starting to be a bit paranoïd with macro message, and wanted to cover all cases :)

Sure but it's impossible to do that ;)

Did you also test with a table inside? or a Group Block ? or ... :)

vmassol: In short we can't redo all tests that exist on wiki markup

vmassol:
> Actually I indeed forgot to ask @mflorea about what's would be the impact on CKEditor side. Would it work properly?

For sure since it was like this before :)

vmassol:
> Ending up with inline content directly inside a GroupBlock feels a bit wrong to me. 

Seems it's valid in HTML5 so looks ok to me and a simplification we do in several places (like in table cells, list items, etc.

screenshot 2018-12-05 at 09 34 00

tmortagne:
XDOM is not HTML. Currently GroupBlock is defined as something containing standalone content (we don't have any parser which produce inline content directly under a group block, yes not even the XHTML parser) which makes https://jira.xwiki.org/browse/XRENDERING-544 invalid.

There is two possibilities:
* we change the meaning of GroupBlock and allows inline content in it, in which case this "optimization" have nothing to do in AbstractMessageMacro and should go in AbstractBoxMacro
* we don't change the meaning of GroupBlock, in which case XRENDERING-544 should be closed

vmassol:
Yes, TBH I was hesitating yesterday too about changing the behavior from before for something more standard or leaving it for backward compat reasons. And possibly changing the XHTML renderer to remove the extra `<p>` that is not needed.

Note that I've now modified the LaTeX export code to handle both cases so changing the behavior won't impact it anymore.

If we all agree that we prefer breaking backward compat for something more standard, I'm ok with that.

surli:
> If we all agree that we prefer breaking backward compat for something more standard, I'm ok with that.

I'm not really fond of the change implemented here, for maintaining the backward compat. And as Thomas mentioned if we do it like that in macro-message we should have the same behaviour in other macros.
So I'm ok to break the compat. Maybe it's something we should add to the RN of 10.10 then.

tmortagne:
Closing since it's not really valid.

PR #169, opened 2019-01-17
XCOMMONS-1553: Remove XWikiDOMSerializer in favor of HtmlCleaner DomSerializer

surli:
* Fix tests with the appropriate CDATA comments

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-html/src/test/resources/macrohtml7.test
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-html/src/test/resources/macrohtml7.test
@@ -22,14 +22,12 @@ beginMacroMarkerStandalone [html] [] [<script type="text/javascript">
              alert("Hello, World")
    // ]]>
 </script>]
-onRawText [<script type="text/javascript">//<![CDATA[
-
+onRawText [<script type="text/javascript">/*<![CDATA[*/
    
          if (2 > 1)
              alert("Hello, World")
-   // 
-
-//]]></script>] [xhtml/1.0]
+   //
+/*]]>*/</script>] [xhtml/1.0]
 endMacroMarkerStandalone [html] [] [<script type="text/javascript">
    //<![CDATA[
          if (2 > 1)
--- a/xwiki-rendering-macros/xwiki-rendering-macro-html/src/test/resources/macrohtml4.test
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-html/src/test/resources/macrohtml4.test
@@ -41,11 +41,9 @@ onRawText [<p>a <strong>b</strong> c
 </pre>
 <p> Some  <span>  text 
 </span>
-<!--comment-->  another  <script>//<![CDATA[
-
+<!--comment-->  another  <script>/*<![CDATA[*/
 hello  world
-
-//]]></script>
+/*]]>*/</script>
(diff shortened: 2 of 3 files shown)


surli:
@vmassol @Enygma2002 @tmortagne this one is ready to fix jenkins jobs. Those changes needed to be done after https://github.com/xwiki/xwiki-commons/pull/61

xwiki-rendering-macros/xwiki-rendering-macro-html/src/test/resources/macrohtml7.test
```diff
@@ -22,14 +22,12 @@ beginMacroMarkerStandalone [html] [] [<script type="text/javascript">
...
    // ]]>
 </script>]
-onRawText [<script type="text/javascript">//<![CDATA[
-
+onRawText [<script type="text/javascript">/*<![CDATA[*/
    
          if (2 > 1)
              alert("Hello, World")
-   // 
-
```
vmassol: is the removal of this new line ok?

surli: The removal is only about an empty line. The git diff is a bit messy here.

surli: So yes it's ok.

vmassol: Sure, but I was just realizing that the change is not only about changing the comments from `//` to `//*..*/` but there's also the removal of a new line. Do you know why the newline is removed? Does it mean this new line was supposed to be delete before and we had a bug that is now fixed?

surli: It's definitely something we fixed by relying on HtmlCleaner behaviour. If you check the diff of the tests changed yesterday, there was a '\n' deleted anywhere we change the CDATA comments: https://github.com/xwiki/xwiki-commons/commit/0f2174f5ae17a0b7e0c6d8fa9998a772a2f66139#diff-662b1d8baf4677911842867688c409bfL188

tmortagne: Indeed there is various empty lines in this commit. Now it does not change anything from HTML point of view so maybe it's just some new optimization.

vmassol: Actually I see a newline being added (and not removed) at https://github.com/xwiki/xwiki-commons/commit/0f2174f5ae17a0b7e0c6d8fa9998a772a2f66139#diff-662b1d8baf4677911842867688c409bfR230

surli: yes but the behaviour is consistent: each time you have a commented CDATA you get one newline after it, and one only. 
Before the behaviour was: 2 newlines after a "//" comment, and zero after a "/* */" comment, apparently.

vmassol: ok just wanted to make sure you analyzed it (since I don't have the time right now to check it more). Applying the PR. Thx

PR #173, opened 2019-02-06
XRENDERING-556: Allow to parse inline editable macro parameters in html result

surli:
## Issue

https://jira.xwiki.org/browse/XRENDERING-556

## Solution

  * Define a new metadata attribute for identifying parameter
  * Define an utility method in AbstractMacro to create a non generated
metadata for a specific parameter
  * Specify that the title block is non generated in BoxMacro
  * Add new tests for both rendering and parsing of BoxMacro
  * Handle the new metadata attribute when setting the content in
XWikiMacroHandler
  * Upgrade Macro Box tests

## Test

peek 06-02-2019 17-13

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/macro/box/AbstractBoxMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/macro/box/AbstractBoxMacro.java
@@ -32,6 +32,7 @@ import org.xwiki.rendering.block.Block;
 import org.xwiki.rendering.block.FormatBlock;
 import org.xwiki.rendering.block.GroupBlock;
 import org.xwiki.rendering.block.ImageBlock;
+import org.xwiki.rendering.block.MetaDataBlock;
 import org.xwiki.rendering.block.NewLineBlock;
 import org.xwiki.rendering.listener.Format;
 import org.xwiki.rendering.listener.reference.ResourceReference;
@@ -277,8 +278,12 @@ public abstract class AbstractBoxMacro<P extends BoxMacroParameters> extends Abs
                 // we add the title, if there is one
                 if (!StringUtils.isEmpty(titleParameter)) {
                     // Don't execute transformations explicitly. They'll be executed on the generated content later on.
-                    ret.addChildren(AbstractBoxMacro.this.contentParser.parse(
-                        titleParameter, context, false, true).getChildren());
+                    List<? extends Block> titleBlock = AbstractBoxMacro.this.contentParser.parse(
+                        titleParameter, context, false, true).getChildren();
+
+                    // put the right metadata around it
+                    ret.addChildren(Collections.singletonList(new MetaDataBlock(titleBlock,
+                        AbstractBoxMacro.this.getNonGeneratedContentMetaData("title"))));
                 }
                 if (titleBlockList != null) {
                     ret.addChildren(titleBlockList);
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
@@ -20,8 +20,6 @@
 
 package org.xwiki.rendering.internal.parser.xhtml.wikimodel;
 
-import org.xwiki.stability.Unstable;
-
 /**
  * Common interface to all XWiki Tag Handler.
  *
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
@@ -85,4 +85,11 @@ public interface XWikiWikiModelHandler
      */
     @Unstable
     String NON_GENERATED_CONTENT_STACK = "nonGeneratedContentStack";
+
+    /**
+     * Stack parameter which record the name of the parameter the current content is dedicated to.
+     * @since 11.1RC1
+     */
+    @Unstable
+    String PARAMETER_CONTENT_NAME = "parameterContentName";
 }
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
@@ -69,27 +69,23 @@ public interface XWikiWikiModelHandler
      * Stack parameter which hold the instances of the MacroInfo encountered.
      * @since 10.10RC1
      */
-    @Unstable
     String MACRO_INFO = "macroInfo";
 
     /**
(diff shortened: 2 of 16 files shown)


surli:
@tmortagne @mflorea this is ready for review: I only tested a simple case there. And I did not checked yet the behaviour with the CKEditor. But I first wanted your opinion before going further.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiModelHandler.java
```diff
@@ -85,4 +85,11 @@
      */
     @Unstable
     String NON_GENERATED_CONTENT_STACK = "nonGeneratedContentStack";
+
+    /**
+     * Stack parameter which record the name of the parameter the current content is dedicated to.
+     * @since 11.1RC1
+     */
+    @Unstable
```
tmortagne: This is not an APi since it's an internal package so Unstable annotation is n not needed. I did not notice one was associated with NON_GENERATED_CONTENT_STACK.

surli: Last changes were pretty huge, no surprise that some things were missing ;) I'll remove them both.

mflorea:
Cool! Looks good to me. One question though: should we allow any content in the title parameter of the Box Macro? If needed, on the CKEditor side I can restrict the content if we use a different type than ``java.util.List< org.xwiki.rendering.block.Block >`` for the title parameter.

surli:
> should we allow any content in the title parameter of the Box Macro?

since the metadata specified that the content is of type `java.util.List< org.xwiki.rendering.block.Block >` I think we should enforce it.

> I can restrict the content

what do you mean exactly by restricting the content?

mflorea:
> since the metadata specifying that the content is of type `java.util.List< org.xwiki.rendering.block.Block >` I think we should enforce it.

Well, my question is whether ``java.util.List< org.xwiki.rendering.block.Block >`` is the right type for the title parameter or not. It's the right type for the macro content for sure because we want to allow any wiki syntax in the Box macro content, but do we want to allow any wiki syntax in the title parameter?

> what do you mean exactly by restricting the content?

How does a Box macro with a table in its title look? What would be the wiki syntax? What about a list in the title?

surli:
> do we want to allow any wiki syntax in the title parameter?

I don't see why not

> > what do you mean exactly by restricting the content?
> How does a Box macro with a table in its title look? What would be the wiki syntax? What about a list in the title?

That could be interesting to restrict the kind of content here, but I'm not sure it would be the case in every parameters, and more importantly I don't see where we draw the line: e.g. I would allow a picture in a box title, but not a table. 

So maybe it might be easier to not restrict the content for now, since eventually all users are able to put any content using the wiki editor.

mflorea:
> That could be interesting to restrict the kind of content here, but I'm not sure it would be the case in every parameters

I'm talking only about the title parameter here, not about all the macro parameters. Each macro parameter that can be edited in-line should be able to specify its type, different than ``java.util.List< org.xwiki.rendering.block.Block >`` if needed. ``java.util.List< org.xwiki.rendering.block.Block >`` is OK as a default type for macro parameters that can be edited in-line, but there should be possible to use a different type if the macro author wants to restrict the content for a particular macro parameter.

mflorea:
> eventually all users are able to put any content using the wiki editor.

Is this really true? Have you tried writing (from the wiki editor) a box macro with a table in its title? Is it possible?

surli:
> Is this really true? Have you tried writing (from the wiki editor) a box macro with a table in its title? Is it possible?

I just tried with latest snapshot here's the result:
peek 12-02-2019 15-33

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/AbstractMacro.java
```diff
@@ -281,4 +281,32 @@ protected MetaData getNonGeneratedContentMetaData()
...
+    @Unstable
+    protected MetaData getNonGeneratedContentMetaData(String parameterName)
+    {
+        MetaData metaData = new MetaData();
+        Type contentType;
+
+        if (this.contentDescriptor != null) {
+            contentType = this.contentDescriptor.getType();
+        } else {
+            contentType = DefaultContentDescriptor.DEFAULT_CONTENT_TYPE;
```
mflorea: This is the part that I don't like: you can't control the type used for macro parameters as you can for macro content. Shouldn't the type be taken from the macro parameter descriptor? Parameters already have a type specified there.

surli: Hm indeed you're right I shouldn't rely on content stuff here. So, just to be sure we're on the same page, you suggest I replace the whole if with something like:

```java
if (this.macroDescriptor != null) {
   contentType = this.macroDescriptor
         .getParameterDescriptorMap()
         .get(parameterName)
         .getParameterType();
} else {
   contentType = DefaultParameterDescriptor.DEFAULT_PARAMETER_TYPE;
}
```

So now the question is what would be the value of `DefaultParameterDescriptor.DEFAULT_PARAMETER_TYPE`?

mflorea: > you suggest I replace the whole if

Only the ``else`` actually. But note that the parameter descriptor has both ``getParameterType()`` and ``getDisplayType()``. Shouldn't you use the later?

Regarding ``DefaultParameterDescriptor.DEFAULT_PARAMETER_TYPE`` I would be OK to keep ``DefaultContentDescriptor.DEFAULT_CONTENT_TYPE``.

surli: > Shouldn't you use the later?

If I follow it's the type given by `@PropertyDisplayType`, I'd need to investigate a bit more on that one I didn't followed the discussions there.

surli:
@mflorea so I now make usage of the parameter display type to create the right metadata. In order to properly test it, I added the `PropertyDisplayType` annotation on the title parameter of the box macro, I think it's appropriate but I let you judge. All tests are passing.

xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/macro/box/AbstractBoxMacro.java
```diff
@@ -277,8 +278,12 @@ public Block build() throws MacroExecutionException
...
                 if (!StringUtils.isEmpty(titleParameter)) {
                     // Don't execute transformations explicitly. They'll be executed on the generated content later on.
-                    ret.addChildren(AbstractBoxMacro.this.contentParser.parse(
-                        titleParameter, context, false, true).getChildren());
+                    List<? extends Block> titleBlock = AbstractBoxMacro.this.contentParser.parse(
+                        titleParameter, context, false, true).getChildren();
+
+                    // put the right metadata around it
+                    ret.addChildren(Collections.singletonList(new MetaDataBlock(titleBlock,
+                        AbstractBoxMacro.this.getNonGeneratedContentMetaData("title"))));
```
mflorea: Do you really need the ``AbstractBoxMacro.this.`` prefix here? It looks like the ``getNonGeneratedContentMetaData`` method is inherited from the base class.

surli: Actually I'm in an inner private class here, see: https://github.com/xwiki/xwiki-rendering/pull/173/files/dca4ddeddcd413aff36e6427212cba6af58b2d03#diff-0ae096925d6afadeedd83855fed76f80R190

mflorea: Right, sorry, I missed that.

PR #174, opened 2019-05-07
XRENDERING-562: Support of redundant non-generated content div

surli:
## Issue

https://jira.xwiki.org/browse/XRENDERING-562

## Solution

The proposed solution is twofolds:
  1. I needed to avoid switching the IgnoreElementRule each time a non-generated content div is found, but to use the state of the rule itself. I had to change its API for that but I guess it's ok since it was unstable. 

  2. I also needed to ensure that only one scanner is created for the content of the macro, to avoid to reset its content to an empty value when going into the second closing non-generated div.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,7 +229,18 @@
                     }
                 },
                 "ignore" : [
-                  // Add more ignores below...
+                  {
+                    "code": "java.method.parameterTypeParameterChanged",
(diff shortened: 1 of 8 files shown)


surli:
@mflorea @tmortagne I'd like to have your opinion about this one: first thing to check is the new tests to ensure you agree with the expectations. Thanks.

tmortagne:
The expect make sense to me yes.

mflorea:
The nested ``data-xwiki-non-generated-content`` DIVs that correspond to the **same macro** will cause problems to the CKEditor. I can't comment on the technical reasons behind this decision, but seeing those duplicated DIVs makes me think there's a problem somewhere..

surli:
@mflorea I commented on the issue to explain why it might happen. I'll need this to bulletproof our code if we accept XWIKI-16389.

PR #175, opened 2019-05-10
XRENDERING-563: Handle metadata rendering in HTML macro

surli:
## Issue

https://jira.xwiki.org/browse/XRENDERING-563

## Solution

Handle the metadata in the custom HtmlMacroXHTMLRenderer the same way we do in AnnotatedXHTMLRenderer.


xwiki-rendering-macros/xwiki-rendering-macro-html/src/main/java/org/xwiki/rendering/internal/macro/html/HTMLMacroXHTMLChainingRenderer.java
```diff
@@ -154,4 +157,37 @@ protected BlockStateChainingListener getBlockState()
...
+        Map<String, String> attributes = new LinkedHashMap<>();
+
+        for (Map.Entry<String, Object> metadataPair : metadata.getMetaData().entrySet()) {
+            attributes.put(XHTMLXWikiGeneratorListener.METADATA_ATTRIBUTE_PREFIX + metadataPair.getKey(),
+                metadataPair.getValue().toString());
+        }
+
+        attributes.put("class", XHTMLXWikiGeneratorListener.METADATA_CONTAINER_CLASS);
+
+        this.getXHTMLWikiPrinter().printXMLStartElement(getMetadataContainerElement(), attributes);
```
tmortagne: We might want to find some way to do that only if the target syntax is annotated html. Probably means something like having several html macro renderers (like we have several html renderers) and choose one based on the rendering context target syntax.

surli: > We might want to find some way to do that only if the target syntax is annotated html.

I'm not entirely sure I agree: it would mean that anyone who writes an html macro that contains a `<div data-xwiki-custom-metadata="value">` would have a different behaviour when viewing or editing. And I'm not sure it's something he/she would want.

OTOH I guess that we should check that context in the implem of the new wikimacrocontent to not add the metadata when we're not in edition.

tmortagne: > I'm not entirely sure I agree: it would mean that anyone who writes an html macro that contains a <div data-xwiki-custom-metadata="value"> would have a different behaviour when viewing or editing. And I'm not sure it's something he/she would want.

I'm not suggesting to remove stuff from the HTML content here, I'm only referring HTML generated from rendering blocks (which is what this code is doing) and have consistent behavior with other HTML renderers. If you directly put `<div data-xwiki-custom-metadata="value">` in your html macro you will always end up with `<div data-xwiki-custom-metadata="value">` whatever is the context.

>  when we're not in edition

Edition is just one use case so that's really not the right criteria which is why I talked about target syntax.

surli: Indeed you're right. I'll check how to do that.

surli:
Everything's passing, however we decrease the mutation score with it, since the custom test suite are not handled:
> Mutation score of 17 is below threshold of 24

I suggest to decrease it in the pom.xml since the new feature is actually tested, and we already have an open issue for it: https://github.com/STAMP-project/pitest-descartes/issues/83

xwiki-rendering-macros/xwiki-rendering-macro-html/src/main/java/org/xwiki/rendering/internal/macro/html/HTMLMacroXHTMLRendererFactory.java
```diff
@@ -19,33 +19,90 @@
...
- * Create XHTML Renderers for the HTML Macro, see {@link HTMLMacroXHTMLRenderer}.
+ * Create Renderers for the HTML Macro.
+ * It computes the renderer to creates based on the given hint. By default it creates a {@link HTMLMacroXHTMLRenderer}.
  *
  * @version $Id$
  * @since 2.0M3
  */
-@Component
-@Named("xhtmlmacro/1.0")
+@Component(hints = {"xhtmlmacro/1.0", "htmlmacro+annotatedxhtml/1.0"})
```
tmortagne: This suggest that you only support xhtml target syntaxes but in view mode the target syntax is almost always html.

surli: I only added support for annotatedhtml, all other syntaxes fallback on xhtmlmacro. We can add more support later with custom renderer if needed. The objective in the first place was to be able to handle metadata in annotatedhtml.

tmortagne: It would be safer to add support for the currently standard html renderer we provide (so just two more) to not end up with a WTF when we finally decide to switch to annotatedhtml/5.0 in the WYSIWYG (HTML 5 is not well supported by CKEditor 4 from what I understood so we have to stay on XHTML for now).

More importantly for the short term the fallback on exception which will happen for every single HTML macro execution in view mode (since it won't find any corresponding renderer because we use html5) is going to make it a lot slower.

surli: > More importantly for the short term the fallback on exception which will happen for every single HTML macro execution in view mode (since it won't find any corresponding renderer because we use html5) is going to make it a lot slower.

Hmmm right I didn't suspect the possible impact on perf. OK I'll add it then.

PR #177, opened 2019-09-11
Update from Weblate

xwikiorgci:


xwiki-rendering-macros/xwiki-rendering-macro-content/src/main/resources/ApplicationResources_fr.properties
```diff
@@ -19,8 +19,8 @@
 # ---------------------------------------------------------------------------
 
 ### Macro Content
-rendering.macro.content.name=Contenu
+rendering.macro.content.name=Contenu [Weblate test debug: Please don't change that, I'll remove it myself]
```
surli: This needs to be removed.

PR #182, opened 2020-03-12
XRENDERING-11: XHTML Renderer should not compact TEXTAREA and some of her XHTML elements

manuelleduc:
Additional tests to avoid regressions.


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/test/resources/xhtml10/specific/paragraph/paragraph.test
```diff
@@ -0,0 +1,9 @@
+.#-----------------------------------------------------
+.input|xhtml/1.0
+.# Test XRENDERING-11
```
surli: You can specify the JIRA ID, but better to explain a bit what you're testing.

tmortagne: Try to avoid the jira ID completly IMO (unless it's not possible to give enough information in the comment to explain what the test is about and you need to link to more context).

vmassol: Yes it's been our practice to avoid JIRA issue in all non-test code. It needs to be self standing and the problem explained.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/test/resources/xhtml10/specific/paragraph/paragraph.test
```diff
@@ -0,0 +1,10 @@
+.#-----------------------------------------------------
+.input|xhtml/1.0
```
vmassol: a better solution is to use `inputexpect`. This will test is both ways and is shorter to write.

vmassol: Right now in your test you don't using xhtml as input.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/test/resources/xhtml10/specific/paragraph/paragraph.test
ClemensRobbenhaar: However this is test code ... well, test data. :)

vmassol: @ClemensRobbenhaar you got me! :) 

So, in the past we said to never reference JIRA issues anywhere in code. Across the years I've seen some developer sometime reference JIRA issue in test code and that's acceptable to me but only provided we also explain the problem and point to the jira issue to see minute details.

WDYT?

ClemensRobbenhaar: No argument against more details than just the JIRA ID from me; just pointing out that it is test code :) 
Sometimes I find it useful to have the JIRA issue in the code, too, to not having to look at the git log to see when the test code was introduced. However if the issue tracker for the project is ever changed, then this information becomes fairly useless.
But then, I guess the discussion here is moot, as the version of the code we are discussing is outdated and already fixed and merged.

PR #183, opened 2020-03-12
XRENDERING-141: Docbook parser fails to support Italic syntax nested inside Bold syntax

manuelleduc:
Removing the failingTests filter since it does not fails anymore.

### DIFF
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-docbook/src/test/resources/docbook44/config.properties
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-docbook/src/test/resources/docbook44/config.properties
@@ -24,8 +24,5 @@
 # http://www.docbook.org/tdg/en/html/literallayout-x.html
 notApplicableTests = .*monospace.*
 
-# See http://jira.xwiki.org/jira/browse/XRENDERING-141
-failingTests = .*bold3.*
-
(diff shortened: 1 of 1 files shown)


vmassol:
@manuelleduc thanks but I see only a single commit removing the test. Does it mean the issue should be closed as "cannot reproduce"?

vmassol:
Also I'm surprised it doesn't fail anymore since I don't remember any work done in this area. Do you know what fixed it?

manuelleduc:
I did not inspect it further but I tried directly on a live instance of XWiki 12.2 and by removing the `failingTests` line on 12.2 and 11.10.3 without error.
I guess it could be fixed by a change in a dependency. I'll look into it.

manuelleduc:
I reproduced the error in xwiki-rendering 5.1 and not in 6.0. That means the issue has been fixed by the transition from doxia 1.4 to doxia 1.5. I'm trying to find a release not for doxia but without much success so far.

vmassol:
> I reproduced the error in xwiki-rendering 5.1 and not in 6.0. That means the issue has been fixed by the transition from doxia 1.4 to doxia 1.5 (in 2014). I'm trying to find a release not for doxia but without much success so far.

I think what you did is good enough. Cool! Thanks for investigating

PR #190, opened 2021-12-22
XRENDERING-632: Figure and figure caption events are not handled by the composite and queue listener

michitux:
Jira issue: https://jira.xwiki.org/browse/XRENDERING-632

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/CompositeListener.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/CompositeListener.java
@@ -335,6 +335,38 @@ public class CompositeListener implements Listener
     }
 
     @Override
+    public void beginFigure(Map<String, String> parameters)
+    {
+        for (Listener listener : this.listeners) {
+            listener.beginFigure(parameters);
+        }
+    }
+
+    @Override
+    public void endFigure(Map<String, String> parameters)
+    {
+        for (Listener listener : this.listeners) {
+            listener.endFigure(parameters);
+        }
+    }
+
+    @Override
+    public void beginFigureCaption(Map<String, String> parameters)
+    {
+        for (Listener listener : this.listeners) {
+            listener.beginFigureCaption(parameters);
+        }
+    }
+
+    @Override
+    public void endFigureCaption(Map<String, String> parameters)
+    {
+        for (Listener listener : this.listeners) {
+            listener.endFigureCaption(parameters);
+        }
+    }
+
+    @Override
     public void endSection(Map<String, String> parameters)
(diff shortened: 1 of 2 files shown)


vmassol:
Thanks. I see no unit tests, isn't this going to reduce the coverage and fail with `-Pquality`? 

I guess they're not used ATM and thus there's no integration tests that will exercise them either.

michitux:
I ran a build with `mvn clean install -Plegacy,integration-tests,quality` locally and it passed. However, I wondered if it wouldn't be a good idea to have unit tests that systematically check if the complete listener API is implemented in all those generic listeners and if these implementations do the right thing. I would need some time to find out if and how this is possible in Java without writing separate code for every method, though.

vmassol:
I'm not sure since for me this is not a bug (the fact that some listeners don't do anything for some events or don't handle these events in any specific manner). It just means there was no need and I like the YAGNI principle. We could consider that it's up to the first use case needing it to implement it.

tmortagne:
> check if the complete listener API is implemented in all those generic listeners

Yes, it would be a good thing to make sure no event has been forgotten in those. One possibility for `CompositeListener` would be to pass a 2 mocked Listener to a new CompositeListener, use reflection to call all Listener declared methods on the CompositeListener and check every time if the corresponding method has been called on the mocks.

michitux:
> I'm not sure since for me this is not a bug (the fact that some listeners don't do anything for some events or don't handle these events in any specific manner). It just means there was no need and I like the YAGNI principle. We could consider that it's up to the first use case needing it to implement it.

I think implementations like `CompositeListener`, `QueueListener`, `AbstractChainingListener` and `LookaheadChainingListener` should always implement the full interface. A user of such a listener shouldn't need to check if the listener implements the required methods. Otherwise, this will lead to very surprising bugs as for example in this `LookaheadChainingListener` implemented the method `beginFigure` but passed it to a `QueueListener` that did not implement it.

> For QueueListener I guess it's not easy to make sure the right thing is done since it's not easy to know what is the right EventType for a given method but at least you could check if calling a method save a new event.

Couldn't we check if the saved event calls the correct method by consuming the events using a mock listener?

tmortagne:
> Couldn't we check if the saved event calls the correct method by consuming the events using a mock listener?

Indeed, good idea, that should do the trick.

vmassol:
> I think implementations like CompositeListener, QueueListener, AbstractChainingListener and LookaheadChainingListener should always implement the full interface

Sounds ok indeed. Thx

michitux:
Opened [XRENDERING-633](https://jira.xwiki.org/browse/XRENDERING-633) to track this.

PR #191, opened 2021-12-22
XRENDERING-629: The figure and figure caption should not be part of the editable content

michitux:
* Put the `FigureBlock` and `FigureCaptionBlock` outside the `MetaDataBlock` that wraps the editable content.
* Mark `FigureCaptionBlock` as container for block-level elements. This is consistent with HTML5 and required to keep the caption editable in CKEditor.
* Add a new test case that contains an image.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-629

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
@@ -94,7 +94,6 @@ public class FigureCaptionMacro extends AbstractNoParameterMacro
         {
             XDOM xdom = this.contentParser.parse(content, context, false, false);
             List<Block> figureCaptionChildren = xdom.getChildren();
-            this.parserUtils.removeTopLevelParagraph(figureCaptionChildren);
 
             // Mark the macro content as being content that has not been transformed (so that it can editable inline)
             figureCaptionChildren = Collections.singletonList(new MetaDataBlock(figureCaptionChildren,
--- a/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
@@ -89,15 +89,17 @@ public class FigureCaptionMacro extends AbstractNoParameterMacro
 
         // If we're not inside a FigureBlock then don't do anything.
         Block parent = context.getCurrentMacroBlock().getParent();
-        if (parent instanceof FigureBlock) {
+        if (parent instanceof FigureBlock
+            || (parent instanceof MetaDataBlock && parent.getParent() instanceof FigureBlock))
+        {
             XDOM xdom = this.contentParser.parse(content, context, false, false);
             List<Block> figureCaptionChildren = xdom.getChildren();
             this.parserUtils.removeTopLevelParagraph(figureCaptionChildren);
-            figureCaptionChildren = Collections.singletonList(new FigureCaptionBlock(figureCaptionChildren));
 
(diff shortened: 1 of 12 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-figure/src/test/resources/macrofigure5.test
```diff
@@ -0,0 +1,49 @@
+.runTransformations:macro
+.#-----------------------------------------------------
+.input|xwiki/2.0
```
vmassol, line 3: please add here a description of this test does. Thx

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureCaptionMacro.java
```diff
@@ -89,15 +89,17 @@ public boolean supportsInlineMode()
 
         // If we're not inside a FigureBlock then don't do anything.
         Block parent = context.getCurrentMacroBlock().getParent();
-        if (parent instanceof FigureBlock) {
+        if (parent instanceof FigureBlock
+            || (parent instanceof MetaDataBlock && parent.getParent() instanceof FigureBlock))
```
vmassol, line 93: Small Note (not needed to accept the PR): I think we have some code elsewhere that also skips some blocks. Would be good to introduce some common API to decide if a parent block is of a given type, and skips blocks such as MetaDataBlock or MacroMarkerBlock.

I can remember code in the LaTeX repo at least but I'm sure there's more elsewhere. One idea is to provide a `getXXXParent()` method that returns an effective parent.

https://github.com/xwiki-contrib/latex/blob/master/latex-syntax/src/main/java/org/xwiki/contrib/latex/internal/DefaultLaTeXTool.java#L149

Needed here too: https://github.com/xwiki-contrib/macro-numbered-references/blob/f361080ce569ecfe280516908fd7abd7b03837dc/src/main/java/org/xwiki/contrib/numberedreferences/internal/NumberedFiguresTransformation.java#L196

Anyway, just a thought.

vmassol:
Looking good. I don't master the changes in the listeners but if all the tests continue to pass (and if inline editing still works fine), it should be good! :) Thx!

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/test/resources/macrofigure5.test
michitux: I have added a description in 4be50bc1db612ed4699d73267188a21895f28675.

michitux:
I did some further (manual) testing and found an issue: As the top-level paragraph is removed by the figure caption macro, the content like text and macros is now a direct child of the figure caption. This is a problem when you place an id macro in the content as the id is rendered as a `<div>`-element when we are not in an inline context. Further, the id macro is also displayed as an empty `<div>`, somehow the placeholder is hidden - potentially a bug in the detection of editable content?

The easiest solution would be to stop removing the wrapping paragraph and thus always have the structure `<figcaption><p>Caption content</p></figcaption>`. This would also mirror what you already get in CKEditor right now as in the browser you always have a `<p>`-tag around the caption content. However, this probably creates issues with the numbered content application extension as it currently relies on the content being plain text and not paragraphs.

michitux:
I have now pushed a solution with an innner paragraph in the figure caption. This is a breaking change for numbered content, the CSS will need to be adjusted, otherwise numbers will be on their own line - and I do not see how this CSS can be backwards-compatible, unfortunately. The XHTML renderer currently outputs a `<p>`-tag around the figure caption, which obviously doesn't work anymore if the content is wrapped in another `<p>`-tag. I see two options:

* Output a `<div>`, possibly with a class. I have implemented this with a `figcaption` class.
* Do not output anything, similar to how figures are handled.

I slightly prefer the first option, which is why I have implemented it, but I think this is something that needs some more input as if we do this I think this could be a generic solution, as we can use such a class not only in the XHTML output, but also, for example in XWiki syntax in case we do not support an element. Further, it could be an easy solution to let WikiModel handle all similar HTML5 tags - handle them like a div (i.e., a document in WikiModel) with a certain class.

PR #194, opened 2022-01-10
XRENDERING-631: The annotated HTML5 renderer should not render figcaption inside macros

michitux:
* Change the annotated HTML5 renderer to only render figure captions when inside a figure.
* Extend BlockStateChainingListener with an event stack to allow getting the parent event.
* Extend BlockStateChainingListenerTest to test for parent events.

As proposed in #191, the renderer could also render `<div class="figcaption">` or a similar tag instead of the `<figcaption>`-tag as it would be nice to have at least some HTML tag in the output to identify the caption (it's just that `<figcaption>` is only allowed as direct child of `<figure>`), see also https://jira.xwiki.org/browse/XRENDERING-625 - I haven't included it here as I haven't received any feedback on the proposed `<div>`-tag in #191 yet.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-631


michitux:
I have just pushed another commit that refactors the unit tests to use parameterized tests instead of dynamic tests as they seem more fitting for this use case. Further, this allows to extract the parameter providers to implement https://jira.xwiki.org/browse/XRENDERING-633 without duplicating the code for getting the methods and suitable parameters. We could even extend the parameter providers later to test several parameter values on a single method (like: all heading levels, inline and standalone verbatim, ...).

PR #195, opened 2022-01-12
XRENDERING-375: Add a parser for HTML 5/XHTML 5

michitux:
* Add WikiModel events for figure and figure captions.
* Add a (basic) parser for HTML 5.
* Add a (basic) parser for XHTML 5.
* Handle the figure and figcaption tags.
* Handle span-tags with monospace and verbatim content.
* Add several tests for monospace content.
* Add a test case for parsing figure and figure caption.

Jira issues: https://jira.xwiki.org/browse/XRENDERING-375 / https://jira.xwiki.org/browse/XRENDERING-627 (I think the latter should be closed as duplicate). Also: https://jira.xwiki.org/browse/XRENDERING-506

At the moment, this adds two parsers that both handle HTML 5, while one expects valid XML, the other doesn't. In XWiki, we probably actually need the one that expects XML as we have a custom HTMLCleaner configuration for the WYSIWYG editor's input anyways and thus running HTMLCleaner twice would be a waste of resources. However, as the syntax is named HTML 5 and not XHTML 5 I think we should also have a parser that uses HTMLCleaner internally to accept HTML 5 that is not also XHTML 5.  Should I move the XHTML 5 parser to its own module?


michitux:
Just noticed that the test is failing, this is due to https://github.com/xwiki/xwiki-commons/pull/157 not being merged yet (locally it is passing as it uses my local version of `xwiki-commons`). So this definitely depends on https://github.com/xwiki/xwiki-commons/pull/157 at least for the non-XML parser.

michitux:
I have added a new module `xwiki-rendering-syntax-xhtml5` with a new `xhtml/5` syntax. I have intentionally not used a minor version here. While there exist HTML 5.0, 5.1 and 5.2 standards, the current status is that nobody intents to release another numbered version but rather there is a living standard that is going to be changed while remaining backwards-compatible, see also [What is HTML5](https://github.com/whatwg/html/blob/main/FAQ.md#what-is-html5). I have not changed the existing `html/5.0` syntax as I do not know how to do this while remaining backwards-compatible.

While I believe this is basically ready to be merged, there are a couple of open questions:

* Is creating a separate (X)HTML5 parser the right thing or should we just make the existing (X)HTML parsers compatible with HTML5 tags and make them also available under a version "5" tag? The new parsers should be backwards-compatible with the existing parsers in the sense that all input that is accepted by the old parsers should also be accepted by the (X)HTML5 parsers. If this is not the case I would consider this a bug that should be fixed.
* If we do not decide for this first option, should we still add some compatibility to the old parsers, like parsing `<span class="monospace">` into monospace format just in case somebody accidentally parses content produced by the HTML5 renderer using the old parser?

Further, this is of course far from being a complete parser for HTML5, this is just parsing a small subset of HTML5 that should correspond to the subset of HTML5 that is produced by the corresponding renderers.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml5/src/main/java/org/xwiki/rendering/internal/parser/xhtml5/wikimodel/XWikiSpanTagHandler.java
```diff
@@ -0,0 +1,104 @@
...
+
+import org.xwiki.component.manager.ComponentManager;
+import org.xwiki.rendering.internal.parser.xhtml.XHTMLParser;
+import org.xwiki.rendering.wikimodel.WikiParameter;
+import org.xwiki.rendering.wikimodel.WikiParameters;
+import org.xwiki.rendering.wikimodel.xhtml.handler.TeletypeTagHandler;
+import org.xwiki.rendering.wikimodel.xhtml.impl.TagContext;
+
+/**
+ * XWiki HTML5 span tag handler. Also handles monospace text.
```
tmortagne, line 35: HTML5 -> XHTML5 ?

tmortagne:
> I have not changed the existing html/5.0 syntax as I do not know how to do this while remaining backwards-compatible.

I guess all it takes is for the component associated with hint "html/5.0" to also be associated to "html/5".

michitux:
> I guess all it takes is for the components associated with hint "html/5.0" to also be associated to "html/5".

There is also the `getSyntax()`-method which returns just one syntax. Thus I'm not sure we can do this using just a single class or if we indeed need a legacy class. I'm also not sure if it is worth the effort. The string `html/5.0` is also already used in various templates as output syntax, e.g., for rendering UIXs or also as output syntax of the Flamingo skin and therefore I'm not sure it is a good idea to deprecate it.

tmortagne:
> There is also the getSyntax()-method which returns just one syntax

Yes, but as I said you would have two component instances in practice, so you can dynamically return a different syntax depending on the component instance hint.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/syntax/Syntax.java
```diff
@@ -48,6 +49,14 @@
  */
 public class Syntax implements Comparable<Syntax>
 {
+    /**
+     * XHTML5 syntax.
+     *
+     * @since 14.0RC1
```
tmortagne, line 55: Just need to update the various `@since` and I think we are good to go.

PR #200, opened 2022-01-18
XRENDERING-639: Allow recursive inheritance for the CTS

michitux:
Jira issue: https://jira.xwiki.org/browse/XRENDERING-639

### DIFF
--- a/xwiki-rendering-test/src/main/java/org/xwiki/rendering/test/cts/TestDataParser.java
+++ b/xwiki-rendering-test/src/main/java/org/xwiki/rendering/test/cts/TestDataParser.java
@@ -139,9 +139,13 @@ public class TestDataParser
 
         // If the inherit configuration property is set and if the returned syntax is empty load from the inherit
         // syntax.
-        if (configuration.inheritSyntax != null) {
+        TestDataConfiguration currentConfiguration = configuration;
+        while ((testDataIN.syntaxData == null || testDataOUT.syntaxData == null)
+            && currentConfiguration.inheritSyntax != null) {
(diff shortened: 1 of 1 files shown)


xwiki-rendering-test/src/main/java/org/xwiki/rendering/test/cts/TestDataParser.java
```diff
@@ -151,6 +155,13 @@
                 testDataOUT.syntaxData = inheritedSyntaxData.getRight().getLeft();
                 testDataOUT.syntaxExtension = inheritedSyntaxData.getRight().getRight();
             }
+
+            try {
+                currentConfiguration = parseTestConfiguration(inheritedSyntaxDirectory, ctsRootPackageName,
+                    classLoader);
+            } catch (Exception ignored) {
+                break;
```
vmassol, line 163: Why is this ignored? Maybe add a comment to explain why.

michitux: Good question, I think my initial assumption was somehow that we might get an exception if there is no more syntax to inherit from, but looking again this doesn't make any sense. I think this exception shouldn't be ignored actually so we notice errors with the configuration files. However, then we would need to change the `throws`-clause which would break all callers of this methods from my understanding. I've just tried doing that and it seems revapi is not complaining which really surprises me. So is it okay to change the `throws`-clause from `IOException` to `Exception`?

tmortagne: > I've just tried doing that and it seems revapi is not complaining which really surprises me

We exclude test stuff AFAIK.

PR #206, opened 2022-02-07
XRENDERING-628: Treat images that are alone in a paragraph and have a label as captioned figure

michitux:
* Detect paragraphs with a single image with label and transform them into a figure in the parser.
* Handle figure and figure caption events in the XWiki syntax renderer.
* Introduce a chaining listener to analyze the content of the figure.
* Add tests to test the new syntax.
* Extend parser to allow nested links if it ends with an image reference.
* Transparently add and remove a class="image" parameter which is required by CKEditor.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-628

Tests for this pull requests will fail as this depends on:
* #195 for the integration tests with HTML 5 input.
* #191 for integration tests at least with XHTML 1.0 as it changes the HTML output.
* #204 for `ListenerMethodProvider`

Further, this needs a corresponding change in the CKEditor integration to allow `<figure>` and `<figcaption>`-tags.

I'm posting this nevertheless to get some early feedback. In particular I'm wondering:

* Is this `StackingFigureContentChainingListener` the right approach?
* Is it okay to just queue the whole paragraph's content or should we exit queuing early once more than one item has been added to the queue?
* Would it be a good idea to introduce a helper to simplify removing/adding classes in the parameters?
* This increases the class fan-out complexity of `XWikiSyntaxChainingRenderer` above the limit of 20. It is excluded from checks, anyways, but if you have some suggestions how to improve it, let me know.


xwiki-rendering-wikimodel/src/main/javacc/XWiki20Scanner.jj
```diff
@@ -422,8 +422,13 @@ TOKEN_MGR_DECLS: {
...
     | <#HEADER_BEGIN_PATTERN: (<SPACE>)* ("=")+ (<SPACE>)* >
-    | <#REFERENCE_IMAGE: "[[image:" ("~" ~[] | ~["]", "~"] | "]" ~["]", "~"]  | "]~" ~[] )* "]]" >
-    | <#REFERENCE:  ( "[[" ( <REFERENCE_IMAGE> | "~" ~[] | ~["]", "~"] | "]" ~["]", "~"] | "]~" ~[] )* "]]" ) >
+    | <#REFERENCE_INNER: "~" ~[] | ~["]", "~"] | "]" ~["]", "~"] | "]~" ~[] >
+    | <#REFERENCE_IMAGE_INNER: "image:" (<REFERENCE_INNER>)* >
+    | <#REFERENCE_IMAGE: "[[" <REFERENCE_IMAGE_INNER> "]]" >
+    | <#REFERENCE_SINGLE:  ( "[[" ( <REFERENCE_IMAGE> | <REFERENCE_INNER> )* "]]" ) >
+    // Allow nested references, but only if the reference ends with an image reference
+    | <#REFERENCE:  ( "[[" ( <REFERENCE_IMAGE> | <REFERENCE_INNER> )* (( <REFERENCE_SINGLE> | <REFERENCE_IMAGE> |
+    <REFERENCE_INNER> )* ">>" <REFERENCE_IMAGE_INNER>)? "]]" ) >
```
tmortagne, line 431: Modifying the behavior of the low level parser might be a risk in terms of retro compatibility which usually leads to introducing a new syntax (2.2) but it's not fully clear to me what can be the actual consequence here. Do you have some example ?

michitux:
I have a noticed a potential issue with the stacking figure content listener: I'm basically duplicating the `beginFigure`-event. This seemed like the easiest solution for me, however, it means that for example the block state chaining listener will also receive the event twice. This is for example problematic with the new "parent event" feature I introduced. I could make this feature more robust but I'm wondering if there isn't an alternative like having a custom callback that is called instead of the duplicated `beginFigure`-event. What do you think, is it important not to duplicate `beginFigure`?

PR #209, opened 2022-03-07
XRENDERING-651: Automatically generate ids for images

michitux:
* Generate ids similar to headings in DefaultXWikiGeneratorListener for WikiModel-parsers and XWikiGeneratorSink in the Doxia parser.
* Add a new id attribute to ImageBlock.
* Add a new version with id parameter to ImageListener.onImage.
* Add retro-compatibility support for onImage to AbstractChainingListener.onImage.
* Implement the new onImage method in all generic listeners.
* Output the id in the XHTML renderer and ignore the generated id in the XHTML parser.
* Adapt all integration tests to expect the new generated id.
* Migrate EventsChainingRendererTest to Junit5.
* Add a unit test for wrapping listener.

Some points we might discuss:
* Should we really take the full reference for the id? In particular in the case of base64-encoded images this can generate long ids which may be difficult to use. There seems to be no limit how long an id can be in HTML.
* Is it okay to add the new id parameter before the generic parameters? I did that to make it consistent with the other methods but then again it might be strange to have an optional parameter that is not at the end.
* Should `ImageBlock` call the old `onImage`-method when the id is `null`?
* Should the icon transformation and the box macro generate ids, too? Currently, they generate images without ids.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-651


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ImageBlock.java
```diff
@@ -54,7 +62,7 @@
      */
     public ImageBlock(ResourceReference reference, boolean freestanding)
     {
-        this(reference, freestanding, Collections.<String, String>emptyMap());
+        this(reference, freestanding, null, Collections.emptyMap());
```
manuelleduc, line 65: `Map.of()` is also an option.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ImageBlock.java
```diff
@@ -65,11 +73,26 @@ public ImageBlock(ResourceReference reference, boolean freestanding)
...
     public ImageBlock(ResourceReference reference, boolean freestanding, Map<String, String> parameters)
+    {
+        this(reference, freestanding, null, parameters);
+    }
+
+    /**
+     * @param reference the image reference
+     * @param freestanding indicate if the image syntax is simple a full descriptive syntax (detail depending of the
+     *            syntax)
+     * @param id the (automatically generated) id of the image.
```
manuelleduc, line 84: We usually don't put a dot at the end of `@param` text.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ImageBlock.java
michitux: We also have `Listener.EMPTY_PARAMETERS`, this change is mostly to remove the superfluous type parameters.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ImageBlock.java
```diff
@@ -121,6 +161,7 @@ public boolean equals(Object obj)
 
             builder.append(getReference(), ((ImageBlock) obj).getReference());
             builder.append(isFreeStandingURI(), ((ImageBlock) obj).isFreeStandingURI());
+            builder.append(getId(), ((ImageBlock) obj).getId());
```
manuelleduc, line 164: `equals/hasCode` can be refactored to use \HashCodeBuilder` and `EqualsBuilder`. See https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HEquals2FHashCodeandToStringimplementations

michitux: You mean the wrong `obj instanceof ImageBlock` or do you mean that we should use method chaining instead of working with an explicit instance variable?

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/chaining/AbstractChainingListener.java
```diff
@@ -67,23 +80,39 @@
      */
     public AbstractChainingListener()
     {
+        this.listItemRetroCompatibility = needsRetroCompatibility(method -> method.getName().endsWith("ListItem"), 0);
+        this.imageRetroCompatibility = needsRetroCompatibility(method -> method.getName().equals("onImage"), 3);
+    }
+
+    /**
+     * @param methodFilter A predicate to filter the methods.
```
manuelleduc, line 88: param text usually start with a lower case.

I've been pointed to https://blog.joda.org/2012/11/javadoc-coding-standards.html for our javadoc style (a long time ago, I'm not sure it's actually in our documentation).

>  **Treat @param and @return as a phrase**
> 
> The @param and @return should be treated as phrases rather than complete sentences. They should start with a lower case letter, typically using the word "the". They should not end with a dot. This aids readability in source code and when generated.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/ImageBlock.java
manuelleduc: ok, should have checked first. This is simply not using the standard `equals + EqualsBuilder` from Intellij. 
The type of the parameters is tested and the methods stopped early if it is not the case, for instance:
```
@Override
public boolean equals(Object obj)
{
    if (this == obj) {
        return true;
    }

    if (obj == null || getClass() != obj.getClass()) {
        return false;
    }

    ImageBlock that = (ImageBlock) obj;

    return new EqualsBuilder().appendSuper(super.equals(obj))
        .append(this.freestanding, that.freestanding).append(this.reference, that.reference).isEquals();
}
```

manuelleduc: This is a small perf improvement (and code style improvement as a bunch of repeated casts can be avoided.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/chaining/AbstractChainingListener.java
michitux: [Our documentation](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HJavadocBestPractices) mentions the [Oracle Javadoc coding conventions](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide) which mentions phrases but doesn't require them. I think I got somehow too much into writing comments as sentences as we prefer them for the initial sentence and also for inline comments. I've fixed this in another commit, I hope I found all of them.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/image/AbstractXHTMLImageTypeRenderer.java
```diff
@@ -84,7 +93,19 @@ public XHTMLWikiPrinter getXHTMLWikiPrinter()
     @Override
     public void onImage(ResourceReference reference, boolean freestanding, Map<String, String> parameters)
     {
-        Map<String, String> attributes = new LinkedHashMap<String, String>();
+        onImage(reference, freestanding, null, parameters);
+    }
+
+    /**
```
tmortagne, line 99: This comment is not needed (the `@Override` is enough).

tmortagne (approved):

PR #211, opened 2022-03-17
XRENDERING-6: Id are not unique when included document or macro content has the same headers as the top document

michitux:
* Ensure that ids are unique also when nested in links or macros.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-6


tmortagne (approved):

manuelleduc (approved):
The change looks good and definitely improves things.

Note: If we need to pass around some class by parameters again, we'll need to use another pattern (e.g., a context) to propagate shared states.

PR #212, opened 2022-05-05
XRENDERING-663: Restrict allowed attributes in HTML rendering

michitux:
* Change HTML renderers to only print allowed attributes
* Adapt tests to not use invalid attributes

This depends on xwiki/xwiki-commons#170.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-663


tmortagne:
I'm afraid that totally forbidding some HTML elements even in context where it's not a threat will cause quite a few retro compatibility problems.

A pity renderers don't support input properties like filters to at least have a way to enable/disable the sanitizer depending on the context. I guess we could use the RenderingContext for that (temporarily disable the sanitizer or set a custom sanitizer in the RenderingContext for the renderer to use).

michitux:
I've added a feature to preserve disallowed attributes as attributes with prefix `data-xwiki-translated-attribute-` such that attributes like `latex-template` that are used by the [LaTeX Syntax](https://extensions.xwiki.org/xwiki/bin/view/Extension/LaTeX/Syntax%201.0/#HSpecificCustomTemplates) aren't removed when using the WYSIWYG editor and rebased this PR on top of #217 as otherwise the prefixed attributes aren't preserved.

PR #215, opened 2022-06-15
XRENDERING-655: Add the ability to style image captions

manuelleduc:
- Introduce the Hidden macro type
- Introduce the FigureTypeRecognizerMacro
- Use FigureTypeRecognizerMacro in FigureMacro
- Test updated in xwiki-macro-rendering-figure and xwiki-rendering-syntax-html5

Jira: https://jira.xwiki.org/browse/XRENDERING-665

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureTypeRecognizerMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureTypeRecognizerMacro.java
@@ -20,6 +20,7 @@
 package org.xwiki.rendering.internal.macro.figure;
 
 import java.util.List;
+import java.util.Set;
 
 import javax.inject.Inject;
 import javax.inject.Named;
@@ -37,13 +38,15 @@ import org.xwiki.rendering.transformation.MacroTransformationContext;
  * insert a {@code data-xwiki-rendering-figure-type} property with its type in the {@link FigureBlock}.
  *
  * @version $Id$
- * @since 14.5RC1
+ * @since 14.6RC1
  */
 @Component
 @Named("figureTypeRecognizer")
 @Singleton
 public class FigureTypeRecognizerMacro extends AbstractNoParameterMacro
 {
+    private static final String DATA_XWIKI_RENDERING_FIGURE_TYPE = "data-xwiki-rendering-figure-type";
+
     @Inject
     private FigureTypeRecognizer figureTypeRecognizer;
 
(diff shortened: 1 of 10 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureTypeRecognizerMacro.java
```diff
@@ -0,0 +1,77 @@
...
+    private FigureTypeRecognizer figureTypeRecognizer;
+
+    /**
+     * Default constructor.
+     */
+    public FigureTypeRecognizerMacro()
+    {
+        super("Figure Type Recognizer", "Internal macro used to recognizer the type of a figure.");
+        setDefaultCategory(DEFAULT_CATEGORY_HIDDEN);
+        // 1 more than the Figure macro priority (so executed after the Figure macro).
```
michitux, line 57: Shouldn't it have a much higher priority such that it is executed after, e.g., nested script macros that could generate the nested table? Also, isn't the point rather to execute it after the figure caption macro than to execute it after the figure macro?

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureTypeRecognizerMacro.java
```diff
@@ -0,0 +1,77 @@
...
+        return false;
+    }
+
+    @Override
+    public List<Block> execute(Object parameters, String content, MacroTransformationContext context)
+    {
+        Block nextSibling = context.getCurrentMacroBlock().getNextSibling();
+        if (nextSibling instanceof FigureBlock) {
+            String type = this.figureTypeRecognizer.isTable((FigureBlock) nextSibling) ? "table" : "figure";
+            nextSibling.setParameter("data-xwiki-rendering-figure-type", type);
```
michitux, line 73: I think it might be a good idea to first check if the macro already has this attribute. There is currently no way to set it via the figure macro, but it could still be set in some other way.

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureTypeRecognizerMacro.java
```diff
@@ -0,0 +1,77 @@
...
+{
+    @Inject
+    private FigureTypeRecognizer figureTypeRecognizer;
+
+    /**
+     * Default constructor.
+     */
+    public FigureTypeRecognizerMacro()
+    {
+        super("Figure Type Recognizer", "Internal macro used to recognizer the type of a figure.");
```
michitux, line 55: Typo: "recognizer" -> "recognize".

michitux (changes requested):
Apart from the few comments I added this looks good in my opinion. Now it would probably be nicer to have support for hiding macros first, not sure if it is a good idea to introduce that category here.

PR #216, opened 2022-06-20
XRENDERING-664: Allow macros to have several categories

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-664

Note: There is still some ongoing discussion regarding the naming of "categories", see https://forum.xwiki.org/t/add-concept-of-hidden-macros/9903/10

See https://github.com/xwiki/xwiki-platform/pull/1865 for the xwiki-platform part of this work.

### DIFF
--- a/xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/pom.xml
+++ b/xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/pom.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.xwiki.rendering</groupId>
+    <artifactId>xwiki-rendering-legacy-transformations</artifactId>
+    <version>14.5-SNAPSHOT</version>
+  </parent>
+  <artifactId>xwiki-rendering-legacy-transformation-macro</artifactId>
+  <name>XWiki Rendering - Legacy - Transformation - Macro</name>
+  <packaging>jar</packaging>
+  <description>Legacy module for xwiki-rendering-transformation-macro</description>
+  <properties>
+    <xwiki.jacoco.instructionRatio>0.00</xwiki.jacoco.instructionRatio>
+    <!-- The features provided by this module so that it's found when resolving extension. -->
+    <xwiki.extension.features>org.xwiki.rendering:xwiki-rendering-transformation-macro</xwiki.extension.features>
+  </properties>
+  <dependencies>
+    <!-- Trigger xwiki-rendering-transformation-macro dependencies (but without xwiki-rendering-transformation-macro jar itself). -->
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-transformation-macro</artifactId>
+      <version>${project.version}</version>
+      <type>pom</type>
+    </dependency>
+    <!-- We need this dependency so that the wrapped module is built before this one. -->
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-transformation-macro</artifactId>
+      <version>${project.version}</version>
+      <!-- We don't want to draw this dependency since we're wrapping it. -->
+      <scope>provided</scope>
+    </dependency>
+    <!-- Testing Dependencies. -->
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-tool-test-component</artifactId>
+      <version>${commons.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+      <!-- Apply Compatibility Aspects. -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>aspectj-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>backward-compatibility-aspects</id>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <weaveDependencies>
+            <weaveDependency>
+              <groupId>org.xwiki.rendering</groupId>
+              <artifactId>xwiki-rendering-transformation-macro</artifactId>
+            </weaveDependency>
+          </weaveDependencies>
+        </configuration>
+      </plugin>
+      <!-- Exclude AspectJ's builddef.lst file form the generated JAR since it's not useful there. -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <excludes>
+            <exclude>**/builddef.lst</exclude>
+          </excludes>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <classpathDependencyExcludes>
+            <classpathDependencyExcludes>org.xwiki.rendering:xwiki-rendering-transformation-macro:jar</classpathDependencyExcludes>
+          </classpathDependencyExcludes>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>fr.inria.gforge.spoon</groupId>
+        <artifactId>spoon-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>spoon-test</id>
+            <configuration>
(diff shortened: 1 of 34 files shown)


xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/src/main/aspect/org/xwiki/rendering/macro/descriptor/AbstractMacroDescriptorAspect.aj
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @version $Id$
+ * @since 14.6RC1
+ */
+public privileged aspect AbstractMacroDescriptorAspect
+{
+    /**
+     * @param defaultCategory default category under which this macro should be listed
+     * @see MacroDescriptor#getDefaultCategories()
+     * @deprecated since 14.6RC1 use {@link AbstractMacroDescriptor#setDefaultCategories(String...)} or {@link  AbstractMacroDescriptor#setDefaultCategories(Set)}
```
vmassol, line 16: Stupid question: what is a non-default category?

xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/src/test/java/org/xwiki/rendering/macro/descriptor/DeprecatedAbstractMacroDescriptorTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @Test
+    void setDefaultCategory()
+    {
+        AbstractMacroDescriptor abstractMacroDescriptor = new AbstractMacroDescriptor(new MacroId("testmacro"), "testmacro", "description",
+            new DefaultContentDescriptor("description", false), mock(BeanDescriptor.class))
+        {
+        };
+
+        abstractMacroDescriptor.setDefaultCategory("testcategory");
+        assertEquals(Set.of("testcategory"), abstractMacroDescriptor.getDefaultCategories());
```
vmassol, line 48: That looks like a possible flicker. You should use hamcrest API IMO.

xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    <groupId>org.xwiki.rendering</groupId>
+    <artifactId>xwiki-rendering-legacy-transformations</artifactId>
+    <version>14.6-SNAPSHOT</version>
+  </parent>
+  <artifactId>xwiki-rendering-legacy-transformation-macro</artifactId>
+  <name>XWiki Rendering - Legacy - Transformation - Macro</name>
+  <packaging>jar</packaging>
+  <description>Legacy module for xwiki-rendering-transformation-macro</description>
+  <properties>
+    <xwiki.jacoco.instructionRatio>0.00</xwiki.jacoco.instructionRatio>
```
vmassol, line 35: 0% but you have declared some test dep below. If there are tests it shouldn't be 0, no?

xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    <!-- The features provided by this module so that it's found when resolving extension. -->
+    <xwiki.extension.features>org.xwiki.rendering:xwiki-rendering-transformation-macro</xwiki.extension.features>
```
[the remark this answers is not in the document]
manuelleduc: The corresponding PR https://github.com/xwiki/xwiki-platform/pull/1865

xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/src/main/aspect/org/xwiki/rendering/macro/descriptor/AbstractMacroDescriptorAspect.aj
manuelleduc: This is a list of categories before they are overridden. The non-default categories can be defined using `xwiki.properties` or the Admin UI. 
Note that while `org.xwiki.rendering.internal.macro.DefaultMacroCategoryManager#getMacroIdsByCategory` takes the override into account, to my knowledge no code visible to the end users currently take the overriding into account.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/AbstractMacro.java
```diff
@@ -241,20 +245,50 @@ protected void setDescriptor(MacroDescriptor descriptor)
...
 
+    /**
+     * Allows sub-classes to the default macro categories. This method only has an effect of the internal
+     * {@link MacroDescriptor} is of type {@link AbstractMacroDescriptor}.
+     *
+     * @param defaultCategories the default macro categories to set
+     * @since 14.6RC1
+     */
+    @Unstable
+    protected void setDefaultCategories(String... defaultCategories)
```
vmassol, line 287: why 2 SPIs that do the same thing?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/AbstractMacroDescriptor.java
```diff
@@ -181,16 +185,34 @@ public Map<String, ParameterDescriptor> getParameterDescriptorMap()
     @Override
     public String getDefaultCategory()
     {
-        return this.defaultCategory;
+        return this.defaultCategories.stream().findFirst().orElse(null);
```
vmassol, line 188: There's a problem here. It's a Set and thus has no order and you're returning the first item which could be different on different JVMs (unless you're using a special set). At the minimum it requires a comment to explain why it's going to work.

vmassol: Also, shouldn't this method be moved to legacy?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/MacroDescriptor.java
```diff
@@ -79,9 +80,27 @@
...
      */
+    @Deprecated(since = "14.6RC1")
+    // TODO: move to legacy once cleaned-up from xwiki-platform.
     String getDefaultCategory();
 
+    /**
+     * A macro define a set of classification categories under which it falls. For instance, the "skype" macro would
+     * fall under the "Communication" and "Video" categories. However, a wiki administrator has the ability to override
+     * the default categories for a given macro in order to organize categories as he sees fit. Thus, these default
+     * categories are only an indication from the macro author about what categories the macro should fall.
```
vmassol, line 93: should fall -> should fall under

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/MacroDescriptor.java
```diff
@@ -79,9 +80,27 @@
...
+     *
+     * @return the default categories under which this macro should be listed, of the empty list of the macro does not
```
vmassol, line 95: of the -> or the

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/AbstractMacro.java
```diff
@@ -241,20 +245,50 @@ protected void setDescriptor(MacroDescriptor descriptor)
...
+
+    /**
+     * Allows sub-classes to set default macro categories. This method only has an effect of the internal
+     * {@link MacroDescriptor} is of type {@link AbstractMacroDescriptor}.
+     *
+     * @param defaultCategories the default macro categories to set
+     * @since 14.6RC1
+     */
+    @Unstable
+    protected void setDefaultCategories(Set<String> defaultCategories)
```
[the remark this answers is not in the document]
manuelleduc: I'm providing `protected void setDefaultCategories(Set<String> defaultCategories)` and `protected void setDefaultCategories(String... defaultCategories)` to allow callers to either setDefaultCategories `setDefaultCategories(aSet)` or `setDefaultCategories("categoryA", "categoryB")` without having to initialize a set first.
We could remove the varargs version, it just makes the API slightly easier to use in some cases.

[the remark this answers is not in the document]
manuelleduc: Nothing to do with the aspects, it is protected because it only need to be called from the children afaiu.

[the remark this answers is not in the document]
vmassol: I personally prefer a single signature.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/descriptor/AbstractMacroDescriptor.java
manuelleduc: It will, in a second step as it is used by `xwiki-platform` for now. I've added a todo here: https://github.com/xwiki/xwiki-rendering/pull/216/files#diff-09c9906c170a9fb45c33831886d7206217b9c49fa34d7c107c6335db0648e1b7R86

xwiki-rendering-legacy/xwiki-rendering-legacy-transformations/xwiki-rendering-legacy-transformation-macro/src/test/java/org/xwiki/rendering/macro/descriptor/DeprecatedAbstractMacroDescriptorTest.java
vmassol: > Even if there were several, the order is not supposed to be taken into account when you compare two Set objects.

AFAIK there's no support for that in junit5's assertEquals. Unless you know for a fact that it's supported.

vmassol: Actually it seems you're right and that the Set's equals method support this OOB: https://docs.oracle.com/javase/1.5.0/docs/api/java/util/Set.html#equals%28java.lang.Object%29

Didn't know that.

PR #218, opened 2022-07-01
XRENDERING-672: Provide a way to list the categories of a macro

manuelleduc:
Introduce `MacroCategoryManager#getMacroCategories(org.xwiki.rendering.macro.MacroId)` and its implementation `DefaultMacroCategoryManager#getMacroCategories(org.xwiki.rendering.macro.MacroId)`

Jira: https://jira.xwiki.org/browse/XRENDERING-672

### DIFF
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManager.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManager.java
@@ -19,20 +19,20 @@
  */
 package org.xwiki.rendering.internal.macro;
 
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
 import org.slf4j.Logger;
 import org.xwiki.component.annotation.Component;
+import org.xwiki.properties.ConverterManager;
 import org.xwiki.rendering.macro.Macro;
 import org.xwiki.rendering.macro.MacroCategoryManager;
 import org.xwiki.rendering.macro.MacroId;
@@ -53,6 +53,9 @@ import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMess
 @Singleton
 public class DefaultMacroCategoryManager implements MacroCategoryManager
 {
+    @Inject
+    private ConverterManager converterManager;
+
     /**
      * Used to get macro categories defined by the user (if any).
      */
@@ -187,8 +190,8 @@ public class DefaultMacroCategoryManager implements MacroCategoryManager
         result.put(macroCategory, ids);
     }
 
-    private Set<String> splitCategories(String s)
+    private Set<String> splitCategories(String categories)
     {
-        return Arrays.stream(s.split("\\s*,\\s*")).collect(Collectors.toSet());
+        return new HashSet<>(this.converterManager.getConverter(List.class).convert(List.class, categories));
     }
 }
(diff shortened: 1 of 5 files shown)


xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManager.java
```diff
@@ -161,4 +186,9 @@ private void addToCategory(Map<String, Set<MacroId>> result, MacroId macroId, St
         ids.add(macroId);
         result.put(macroCategory, ids);
     }
+
+    private Set<String> splitCategories(String s)
+    {
+        return Arrays.stream(s.split("\\s*,\\s*")).collect(Collectors.toSet());
```
michitux, line 192: If you really want/need to use a regular expression here, this expression should be pre-compiled as a static variable (this should also produce a warning in SonarLint). Further, why aren't you using our standard list parsing implementation that we normally use for configuration variables?

Also, I think this discards the order of the categories (or at least doesn't guarantee it), is this intentional?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/test/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManagerTest.java
```diff
@@ -19,122 +19,181 @@
...
-import org.xwiki.rendering.syntax.Syntax;
 import org.xwiki.rendering.transformation.macro.MacroTransformationConfiguration;
-import org.xwiki.test.jmock.AbstractComponentTestCase;
+import org.xwiki.test.LogLevel;
+import org.xwiki.test.junit5.LogCaptureExtension;
+import org.xwiki.test.junit5.mockito.ComponentTest;
+import org.xwiki.test.junit5.mockito.InjectMockComponents;
+import org.xwiki.test.junit5.mockito.MockComponent;
+
+import ch.qos.logback.classic.Level;
```
michitux, line 41: Is this really the correct import?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/test/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManagerTest.java
```diff
@@ -19,122 +19,181 @@
...
+        when(this.macroManager.getMacroIds()).thenReturn(Set.of(macroAId));
+        when(this.macroManager.getMacro(macroAId)).thenReturn(macroA);
+        when(macroA.getDescriptor()).thenReturn(macroADescriptor);
 
-        // Check for overwritten categories.
-        Assert.assertTrue(macroCategories.contains("Content"));
-        Assert.assertTrue(macroCategories.contains("Simple"));
+        // Check whether our macros are registered under correct categories.
+        Set<Object> expected = new HashSet<>();
+        expected.add(null);
```
michitux, line 100: Why do we expect a set with `null`? Wouldn't an empty set be better?

michitux (changes requested):

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/test/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManagerTest.java
manuelleduc: Yes, that's the type returned by `ch.qos.logback.classic.spi.ILoggingEvent#getLevel` (`ch.qos.logback.classic.Level`)

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/test/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManagerTest.java
manuelleduc: I've kept the logic where a macro without define category was affected to the "null" category. Without this special category, macros without category wouldn't be returned by some higher level operations.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/macro/DefaultMacroCategoryManager.java
manuelleduc: Sonar is not raising an error (at least not in my ide), but indeed per-compilling the regex is a good idea.
`MacroTransformationConfiguration#getCategories` returns a `java.util.Properties` and not a `commons-configuration` Properties object. And it does not have the notion of list.

PR #219, opened 2022-07-12
XRENDERING-675: Add events to raw content filtering in macros

michitux:
This is a draft implementation of adding events to raw content filtering. In fact, it contains two implementations. I'm not very happy with where the HTML filter now is, I'm open for suggestions for a better location. Maybe in the XHTML renderer? This also adds a "restricted" parameter to the HTML macro to manually control restricted filtering, I'm not sure if this is useful.

Jira issue: https://jira.xwiki.org/browse/XWIKI-18568


xwiki-rendering-macros/xwiki-rendering-macro-html/src/main/java/org/xwiki/rendering/internal/macro/html/HTMLMacro.java
```diff
@@ -152,83 +139,29 @@ public List<Block> execute(HTMLMacroParameters parameters, String content, Macro
...
+                RawBlockFilterParameters filterParameters = new RawBlockFilterParameters();
+                filterParameters.setClean(parameters.getClean());
+                filterParameters.setRestricted(parameters.getRestricted());
+                filterParameters.setMacroTransformationContext(context);
+                for (RawBlockFilter filter
+                    : this.componentManager.<RawBlockFilter>getInstanceList(RawBlockFilter.class)) {
+                    contentBlock = filter.filter(contentBlock, filterParameters);
+                }
+            } catch (ComponentLookupException e) {
+                throw new MacroExecutionException("Couldn't initialize the HTML filtering.", e);
```
manuelleduc, line 154: This block looks duplicated with https://github.com/xwiki/xwiki-rendering/pull/219/files#diff-62bc96d84833c84f3d85d8f3b8f5fe40eb9b8dece414df58660f576f2c6cc8ecR91-R101

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/transformation/macro/RawBlockFilter.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.rendering.block.RawBlock;
+import org.xwiki.rendering.macro.MacroExecutionException;
+
+/**
+ * Filtering for raw blocks to clean and sanitize raw content.
+ *
+ * @version $Id$
+ * @since 14.6RC1
```
manuelleduc, line 30: The target will probably be 14.7RC1?

michitux: Yes, I originally targeted 14.6 but I fear that's too risky.

manuelleduc: With the bank holidays, we are virtually the day before the delivery in terms of working days. So definitely +1 to postpone this for the next RC.

manuelleduc:
@michitux you also probably need to create a Rendering specific issue for the part of you change impacting rendering.

michitux:
> @michitux you also probably need to create a Rendering specific issue for the part of you change impacting rendering.

Done, [XRENDERING-675](https://jira.xwiki.org/browse/XRENDERING-675) is the issue. I'll rebase the commits to change the commit message accordingly.

michitux:
I've added tests and did a few minor fixes in comments and messages. I'm still not fully happy where the HTML filter is but I also found that adding it to the HTML rendering would add a new dependency to the macro transformation which is probably not really desirable even though I think it would be good if each renderer defined its own filter.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/transformation/macro/RawBlockFilterParameters.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * If cleaning of the input should be performed.
+     */
+    private boolean clean;
+
+    /**
+     * The macro transformation context.
+     */
+    private MacroTransformationContext macroTransformationContext;
```
surli, line 49: Looking at the code of `HTMLRawBlockFilter` it looks like this field should never be null. So I'd make it final.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/HTMLRawBlockFilter.java
```diff
@@ -0,0 +1,171 @@
...
+        MacroTransformationContext macroTransformationContext = parameters.getMacroTransformationContext();
+
+        boolean restricted = parameters.isRestricted() || (macroTransformationContext != null
+            && macroTransformationContext.getTransformationContext().isRestricted());
+
+        if (SyntaxType.HTML_FAMILY_TYPES.contains(block.getSyntax().getType())
+            && (parameters.isClean()) || restricted)
+        {
+            String cleanedContent =
+                cleanHTML(block.getRawContent(), parameters.getMacroTransformationContext(), restricted,
```
surli, line 76: You could use directly `macroTransformationContext` here

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/HTMLRawBlockFilter.java
```diff
@@ -0,0 +1,171 @@
...
+
+        // Don't print the XML declaration nor the XHTML DocType.
+        cleanedContent = HTMLUtils.toString(document, true, true);
+
+        // Don't print the top level html element (which is always present and at the same location
+        // since it's been normalized by the HTML cleaner)
+        // Note: we trim the first 7 characters since they correspond to a leading new line (generated by
+        // XMLUtils.toString() since the doctype is printed on a line by itself followed by a new line) +
+        // the 6 chars from "<html>".
+        cleanedContent = cleanedContent.substring(7, cleanedContent.length() - 8);
```
surli, line 144: Sounds a bit dangerous on the long term: shouldn't we check that the cleanedContent does start with `<html>\n` before performing this? and maybe throwing an exception if the content is not exactly the same? I'm a bit afraid that we might break everything if for example a lib change its behaviour to not use a new line anymore.

xwiki-rendering-macros/xwiki-rendering-macro-html/src/main/java/org/xwiki/rendering/internal/macro/html/HTMLMacro.java
```diff
@@ -152,83 +139,28 @@ public List<Block> execute(HTMLMacroParameters parameters, String content, Macro
...
-                    "The HTML macro may not be used with clean=\"false\" in this context.");
+            RawBlock contentBlock = new RawBlock(normalizedContent, targetSyntax);
+
+            try {
+                RawBlockFilterParameters filterParameters = new RawBlockFilterParameters();
+                filterParameters.setClean(parameters.getClean());
+                filterParameters.setMacroTransformationContext(context);
+                for (RawBlockFilter filter
+                    : this.componentManager.<RawBlockFilter>getInstanceList(RawBlockFilter.class)) {
+                    contentBlock = filter.filter(contentBlock, filterParameters);
```
surli, line 150: So it means each filter will perform a filtering on a block that was already filtered previously. So looks like we should introduce a mechanism to order the filters. For example, a filter could transform some attribute that could be removed by another one if they're not transformed.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/transformation/macro/RawBlockFilter.java
```diff
@@ -0,0 +1,53 @@
...
+import org.xwiki.rendering.block.RawBlock;
+import org.xwiki.rendering.macro.MacroExecutionException;
+
+/**
+ * Filtering for raw blocks to clean and sanitize raw content.
+ *
+ * @version $Id$
+ * @since 14.7RC1
+ */
+@Role
```
surli, line 32: Missing `@Unstable`

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/HTMLRawBlockFilter.java
michitux: Yes, I also don't like this code, I've just copied it from the HTML macro.

xwiki-rendering-macros/xwiki-rendering-macro-html/src/main/java/org/xwiki/rendering/internal/macro/html/HTMLMacro.java
```diff
@@ -152,83 +139,28 @@ public List<Block> execute(HTMLMacroParameters parameters, String content, Macro
...
-            } else if (context.getTransformationContext().isRestricted()) {
-                throw new MacroExecutionException(
-                    "The HTML macro may not be used with clean=\"false\" in this context.");
+            RawBlock contentBlock = new RawBlock(normalizedContent, targetSyntax);
+
+            try {
+                RawBlockFilterParameters filterParameters = new RawBlockFilterParameters(context);
+                filterParameters.setClean(parameters.getClean());
+                List<RawBlockFilter> filters = this.componentManager.getInstanceList(RawBlockFilter.class);
+                filters.sort((filter1, filter2) -> filter1.getPriority() - filter2.getPriority());
```
michitux, line 148: In my opinion, `Comparator.comparingInt` would be more readable here:
```suggestion
                filters.sort(Comparator.comparingInt(RawBlockFilter::getPriority));
```

xwiki-rendering-macros/xwiki-rendering-macro-raw/src/main/java/org/xwiki/rendering/internal/macro/raw/RawMacro.java
```diff
@@ -75,8 +84,21 @@ public boolean supportsInlineMode()
...
     @Override
     public List<Block> execute(RawMacroParameters parameters, String content, MacroTransformationContext context)
+        throws MacroExecutionException
     {
         RawBlock rawBlock = new RawBlock(content, parameters.getSyntax());
+
+        try {
+            RawBlockFilterParameters filterParameters = new RawBlockFilterParameters(context);
+            List<RawBlockFilter> filters = this.componentManager.getInstanceList(RawBlockFilter.class);
+            filters.sort((filter1, filter2) -> filter1.getPriority() - filter2.getPriority());
```
michitux, line 94: Same as in the HTML macro: `Comparator.comparingInt` would be more readable. Further, maybe we should indeed remove this code duplication by introducing a helper method, in particular with the sorting step?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/transformation/macro/RawBlockFilterParameters.java
michitux: Yes, though I'm not sure we should really tie this to macro transformations so much. While on the one hand it matches our current use case, we could also extract all properties of the macro transformation we really care about into the parameters and then make the actual filter independent of the macro transformation and thus more generically useful. Depending on what we do with the xdom+xml syntax, we could re-use this filter for example in the xdom+xml parser to filter raw blocks depending on the author. Btw. another use case for this filter (where we have a macro transformation context, though) would be in the context of the RSS macro, where the filtering is currently re-implemented.

xwiki-rendering-macros/xwiki-rendering-macro-html/src/main/java/org/xwiki/rendering/internal/macro/html/HTMLMacro.java
surli: Thanks, didn't know about it.

xwiki-rendering-macros/xwiki-rendering-macro-raw/src/main/java/org/xwiki/rendering/internal/macro/raw/RawMacro.java
surli: > Further, maybe we should indeed remove this code duplication by introducing a helper method, in particular with the sorting step?

I'm actually wondering how much all that shouldn't be a default feature for our components in the future. I think we have already a ticket for the order of components. 

But in the meantime I think I'll indeed extract this so that it's easier to change if we need to.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/HTMLRawBlockFilter.java
surli: Note that I actually changed it a bit in https://github.com/xwiki/xwiki-rendering/pull/219/files#diff-d03e09b8808925132b8ef48ce07ef681ef58c4cb5f28cf00bd0b9284051c2d0dR149.

PR #220, opened 2022-08-19
XRENDERING-676: Introduce the notion of hidden macro categories

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-676

### DIFF
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/DefaultMacroTransformationConfiguration.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/DefaultMacroTransformationConfiguration.java
@@ -43,10 +43,18 @@ public class DefaultMacroTransformationConfiguration implements MacroTransformat
      */
     private Properties macroCategories = new Properties();
 
+    private Set<String> hiddenCategories = Set.of();
+
     @Override
     public Properties getCategories()
     {
         return this.macroCategories;
+    }
+
+    @Override
+    public Set<String> getHiddenCategories()
+    {
+        return this.hiddenCategories;
     }
 
     /**
@@ -61,5 +69,16 @@ public class DefaultMacroTransformationConfiguration implements MacroTransformat
         // This method is useful for those using the XWiki Rendering in standalone mode since it allows the rendering
         // to work even without a configuration store.
         this.macroCategories.setProperty(macroId.toString(), String.join(",", categories));
+    }
+
+    /**
+     * Set the hidden categories.
+     *
+     * @param hiddenCategories the set of hidden categories
+     * @since 14.8RC1
+     */
+    public void setHiddenCategories(Set<String> hiddenCategories)
+    {
+        this.hiddenCategories = hiddenCategories;
     }
 }
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/transformation/macro/MacroTransformationConfiguration.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/transformation/macro/MacroTransformationConfiguration.java
@@ -20,15 +20,20 @@
 package org.xwiki.rendering.transformation.macro;
 
 import java.util.Properties;
+import java.util.Set;
 
 import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
 
 /**
  * Configuration properties for the Macro Transformation module.
  * <p>
- * You can override the default values for each of the configuration properties below by defining them in XWiki's
- * global configuration file using a prefix of "rendering.transformation.macro." followed by the property name.
- * For example: <code>rendering.transformation.macro.categories = toc:my Category</code>
+ * You can override the default values for each of the configuration properties below by defining them in XWiki's global
+ * configuration file using a prefix of "rendering.transformation.macro." followed by the property name. For example:
+ * {@code rendering.transformation.macro.categories = toc:my Category}
+ * <p>
+ * You can also override the default set of hidden macros by defining the
+ * {@code rendering.transformation.macro.hiddenCategories} property.
  *
  * @version $Id$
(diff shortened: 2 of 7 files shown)


xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/transformation/macro/MacroTransformationConfiguration.java
```diff
@@ -42,4 +47,14 @@
...
     Properties getCategories();
+
+    /**
+     * @return the set of hidden categories (e.g., "Deprecated", "Internal")
+     * @since 14.8RC1
+     */
+    @Unstable
+    default Set<String> getHiddenCategories()
+    {
+        throw new UnsupportedOperationException();
```
tmortagne, line 58: I would return an empty Set by default, otherwise there is no point in having a default implementation honestly if it's a failing one.

vmassol: I'm not sure since an empty set would be incorrect if there are hidden categories existing. I prefer the exception in this case as it's cleaner.

tmortagne: The point f the default implementation is retro compatibility and since no category were hidden before that new API that's what the default behavior should reflect IMO.

Throwing an exception here make the default implementation totally useless, not implementing it at all would have the exact same effect. It's just lying to revapi basically.

vmassol: > Throwing an exception here make the default implementation totally useless

I don't agree with this statement. The point is to make any extension implementing the interface not fail with NoSuchMethodException when loaded by the CL. Any existing extension will not call the new method obviously since it's not been modified... So no backward compatibility issue here.

vmassol: (ofc if the extension is called by XWiki platform itself, then it needs to implement the new method or it'll fail)

tmortagne: > The point is to make any extension implementing the interface not fail with NoSuchMethodException (not sure it's this exception but something related) when loaded by the CL.

There will be an exception only when the method is called, not when the class is loaded.

vmassol: Thanks for the link, it's well documented there. I was wrong as I thought the linker would check for missing method implementations during CL. So no backward compatibility issue when adding new methods to interfaces (good to know, I think I've been wrong since the beginning when I proposed to use default methods when moving to Java 8 :)). 

So this means that we don't really need to add a default impl when adding a new interface method for which we throw an UnsupportedOperationException exception. See the numerous places where we do that at https://github.com/xwiki/xwiki-platform/search?q=new+UnsupportedOperationException

tmortagne: > So no backward compatibility issue when adding new methods to interfaces

I don't agree with that. It's clearly a retro compatibility when the interface might be implemented by an extension and used by another or XWiki Standard which knows about the new method, the default implementation should try to return something that make sense in such a case (here what make sense is to indicate that there is no hidden macros since that custom implementation of MacroTransformationConfiguration does not have any way to configure hiden macros).

> See the numerous places where we do that at [xwiki/xwiki-platform/search (new UnsupportedOperationException)](https://github.com/xwiki/xwiki-platform/search?q=new+UnsupportedOperationException)

Not all of them are about default implementation of new methods, but those that are are wrong in most cases IMO. Did not notice some people had done it. We should always try to find a not failing default behavior, focused on retro compatibility.

vmassol: > So no backward compatibility issue when adding new methods to interfaces

I meant no *binary* backward compatibility issue.

It's a problem when it's a SPI as mentioned at https://revapi.org/revapi-java/0.26.1/differences.html#java.method.addedToInterface

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/DefaultMacroTransformationConfiguration.java
```diff
@@ -43,12 +43,20 @@ public class DefaultMacroTransformationConfiguration implements MacroTransformat
...
     @Override
     public Properties getCategories()
     {
         return this.macroCategories;
     }
 
+    @Override
+    public Set<String> getHiddenCategories()
+    {
+        return this.hiddenCategories;
```
vmassol, line 57: Isn't there a problem with the terminology? If Hidden categories are categories then surely getCategories should return them? I don't recall our discussion re hidden macros. If we decided to implement them has categories then I don't understand why we need a new API. Thx

vmassol: ah I remember now, sorry. It's just that we can configure which categories are hidden! :)

manuelleduc: The way the API is currently used, it's easier to have to separate Sets, but indeed that's something that could be useful in the future.

PR #222, opened 2022-09-15
XRENDERING-680: Figure type recognition is sensible to empty lines

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-680


xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/DefaultFigureTypeRecognizer.java
```diff
@@ -66,9 +67,9 @@ private List<Block> getBlocksIgnoringMacroMarkerBlocks(List<Block> blocks)
             } else if (block instanceof MetaDataBlock || block instanceof GroupBlock) {
                 // Ignore MetaData/Group blocks since they're not structural
                 results.addAll(getBlocksIgnoringMacroMarkerBlocks(block.getChildren()));
-            } else if (!(block instanceof FigureCaptionBlock)) {
-                // Ignore figure caption blocks since they're special and do not affect whether the content contains
-                // a table or not.
+            } else if (!(block instanceof FigureCaptionBlock || block instanceof EmptyLinesBlock)) {
+                // Ignore figure caption and empty lines blocks since they're special and do not affect whether the
```
vmassol, line 71: I think we could explain better that we're skipping empty lines because they don't represent content (same as GroupBlock) and we still want to recognize a table if there are emptylines in front or after the table.

vmassol (approved):

PR #223, opened 2022-09-16
XRENDERING-681: Add ability to force a figure type (figure or table)

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-681

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
@@ -96,7 +96,7 @@ public class FigureMacro extends AbstractMacro<FigureMacroParameters>
         FigureBlock figureBlock = new FigureBlock(contentBlock);
         Block block;
         // If a type is explicitly defined, it is used directly. Otherwise, we try to infer it from the macro's content.
-        if (!Objects.equals(parameters.getType(), FigureType.AUTOMATIC)) {
+        if (FigureType.AUTOMATIC != parameters.getType()) {
             figureBlock.setParameter(DATA_XWIKI_RENDERING_FIGURE_TYPE, parameters.getType().getName());
             block = figureBlock;
         } else {
--- a/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
@@ -21,7 +21,6 @@ package org.xwiki.rendering.internal.macro.figure;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 
 import javax.inject.Inject;
--- a/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
@@ -21,6 +21,7 @@ package org.xwiki.rendering.internal.macro.figure;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import javax.inject.Inject;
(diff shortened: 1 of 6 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
```diff
@@ -81,16 +86,25 @@ public boolean supportsInlineMode()
...
         List<Block> contentBlock = List.of(new MetaDataBlock(xdom.getChildren(), getNonGeneratedContentMetaData()));
 
-        return List.of(new CompositeBlock(List.of(
-            new MacroBlock("figureTypeRecognizer", Map.of(), false),
-            new FigureBlock(contentBlock))
-        ));
+        FigureBlock figureBlock = new FigureBlock(contentBlock);
+        Block block;
+        // If a type is explicitly defined, it is used directly. Otherwise, we try to infer it from the macro's content.
+        if (parameters.getType() != null && !Objects.equals(parameters.getType(), FigureType.AUTOMATIC)) {
```
vmassol, line 99: I think the check for null is not needed according to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/macro/figure/FigureMacroParameters.java
```diff
@@ -0,0 +1,60 @@
...
+import org.xwiki.properties.annotation.PropertyDescription;
+import org.xwiki.properties.annotation.PropertyName;
+import org.xwiki.rendering.internal.macro.figure.FigureMacro;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Parameters for the {@link FigureMacro}.
+ *
+ * @version $Id$
+ * @since 14.9RC1
```
vmassol, line 31: Would be better if this could be pushed for 14.8

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/macro/figure/FigureMacroParameters.java
```diff
@@ -0,0 +1,60 @@
...
+    public FigureType getType()
+    {
+        return this.type;
+    }
+
+    /**
+     * @param type the type of the figure (i.e., {@link FigureType#FIGURE} or {@link FigureType#TABLE}}), if
+     *     {@link FigureType#AUTOMATIC} the type will be automatically defined based on the macro content
+     */
+    @PropertyDescription("The type of the figure. When automatic, the type will be defined based on the macro content.")
```
vmassol, line 54: I'd display the different possible values in the desc to make it simpler for users (usability FTW)

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/macro/figure/FigureType.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Explicit table type.
+     */
+    TABLE("table");
+
+    private String name;
+
+    /**
+     * @param name the name of the figure type (e.g., "figure")
```
vmassol, line 52: minor comment to use `{@code figure}`  instead

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/test/resources/macrofigure7.test
```diff
@@ -0,0 +1,66 @@
+.runTransformations:macro
+.#--------------------------------------------------------------------------------------------------
+.input|xwiki/2.0
+.# Test with an explicit figure type.
+.#--------------------------------------------------------------------------------------------------
+{{figure type="FIGURE"}}
```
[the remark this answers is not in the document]
manuelleduc: I've used upper case because that's what is produced by using CKEditor, but +1 to use the lowercase, looks nicer.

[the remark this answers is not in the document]
vmassol: Then that's a problem. We need CKEditor to provide the lowercase IMO, i.e. the canonical value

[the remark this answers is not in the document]
manuelleduc: Looks like currently having lowercases values requires to update `enum/edit.vm` (see https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/html_displayer/enum/edit.vm#L33)
Another option is to defined the entries of FigureType in lowercase.

xwiki-rendering-macros/xwiki-rendering-macro-figure/src/main/java/org/xwiki/rendering/internal/macro/figure/FigureMacro.java
```diff
@@ -81,16 +86,25 @@ public boolean supportsInlineMode()
...
         List<Block> contentBlock = List.of(new MetaDataBlock(xdom.getChildren(), getNonGeneratedContentMetaData()));
 
-        return List.of(new CompositeBlock(List.of(
-            new MacroBlock("figureTypeRecognizer", Map.of(), false),
-            new FigureBlock(contentBlock))
-        ));
+        FigureBlock figureBlock = new FigureBlock(contentBlock);
+        Block block;
+        // If a type is explicitly defined, it is used directly. Otherwise, we try to infer it from the macro's content.
+        if (!Objects.equals(parameters.getType(), FigureType.AUTOMATIC)) {
```
tmortagne, line 99: You don't need to use equals with an enum, == is enough.

tmortagne (approved):

PR #226, opened 2022-10-11
XRENDERING-683: Return a parse exception in case of stackoverflow when parsing with javacc

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-683

### DIFF
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/parser/ParseException.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/parser/ParseException.java
@@ -19,6 +19,8 @@
  */
 package org.xwiki.rendering.parser;
 
+import org.xwiki.stability.Unstable;
+
 /**
  * Encapsulate a parsing error.
  *
@@ -39,6 +41,20 @@ public class ParseException extends Exception
     public ParseException(String message)
     {
         super(message);
+    }
+
+    /**
+     * Constructs a new parse exception with the specified cause .
+     *
+     * @param cause the cause (which is saved for later retrieval by the {@link #getCause()} method).)
+     * @since 14.9RC1
+     * @since 14.4.6
+     * @since 13.10.10
+     */
+    @Unstable
+    public ParseException(Throwable cause)
+    {
+        super(cause);
     }
 
     /**
--- a/xwiki-rendering-api/src/main/java/org/xwiki/rendering/parser/ParseException.java
+++ b/xwiki-rendering-api/src/main/java/org/xwiki/rendering/parser/ParseException.java
@@ -19,8 +19,6 @@
  */
 package org.xwiki.rendering.parser;
 
-import org.xwiki.stability.Unstable;
-
 /**
  * Encapsulate a parsing error.
  *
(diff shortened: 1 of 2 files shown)


xwiki-rendering-api/src/main/java/org/xwiki/rendering/parser/ParseException.java
```diff
@@ -41,6 +43,20 @@ public ParseException(String message)
...
+    /**
+     * Constructs a new parse exception with the specified cause .
+     *
+     * @param cause the cause (which is saved for later retrieval by the {@link #getCause()} method).)
+     * @since 14.9RC1
+     * @since 14.4.6
+     * @since 13.10.10
+     */
+    @Unstable
+    public ParseException(Throwable cause)
```
tmortagne, line 55: I find this constructor very rarely needed, as I feel it's a good practice to always associated a message with the exception.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-wikimodel/src/main/java/org/xwiki/rendering/internal/parser/wikimodel/AbstractWikiModelParser.java
```diff
@@ -91,7 +91,11 @@ protected StreamParser getLinkLabelParser()
     public XDOM parse(Reader source) throws ParseException
```
tmortagne, line 91: I don't think that's the right location. Better make the already existing protection in `parse(Reader source, Listener listener, IdGenerator idGenerator)` cover more cases.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/parser/ParseException.java
manuelleduc: That the first time I have a use for this, because if failed to find a message to add to the `StackOverflowError` catch block. 
Unless we lean for something a bit redundant and generic like "Failed to parse input source"?

tmortagne: Anyway, as I said in another comment the protection already exist somewhere, it just needs to include StackOverflowError so you don't really need that constructor.

manuelleduc: Yep, just realized that :) Fixed

xwiki-rendering-syntaxes/xwiki-rendering-syntax-wikimodel/src/main/java/org/xwiki/rendering/internal/parser/wikimodel/AbstractWikiModelParser.java
```diff
@@ -132,7 +132,9 @@ public void parse(Reader source, Listener listener, IdGenerator idGenerator) thr
         IWikiParser parser = createWikiModelParser();
         try {
             parser.parse(source, createXWikiGeneratorListener(listener, idGenerator));
-        } catch (Exception e) {
+        } catch (Exception | StackOverflowError e) {
+            // Stack overflow exceptions are catch in addition to exceptions because they can be thrown by javacc
```
vmassol, line 136: typo: are catch -> are caught
typo: to many -> too many

PR #230, opened 2022-10-31
XRENDERING-660: Get rid of the dependency on the xdom+xml syntax

michitux:
* Pass the label via an XDOM instance instead of rendering/parsing to xdom+xml syntax.
* Move the reference type detection into the XWikiReferenceTagHandler to simplify XHTMLXWikiGeneratorListener.

I'm not super happy with the new workaround but I think it's better than the old one, avoids the dependency on another syntax and shouldn't cause any breakages as we're (mostly) still doing the same. The better design would be to introduce begin/end methods for references but then we would need (probably) major changes in WikiModel to properly support everything that currently works like having independent styling in link labels.

### DIFF
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiReference.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiReference.java
@@ -49,8 +49,6 @@ public class XWikiWikiReference extends WikiReference
      * @param linkParameters the parameters of the link
      * @param freeStanding if the link is freestanding
      * @since 14.10RC1
-     * @since 14.4.7
-     * @since 13.10.11
      */
     public XWikiWikiReference(ResourceReference reference, XDOM label, WikiParameters linkParameters,
         boolean freeStanding)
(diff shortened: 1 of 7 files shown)


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiReference.java
```diff
@@ -38,12 +39,31 @@ public class XWikiWikiReference extends WikiReference
...
+
+    /**
+     * Construct a new wiki reference.
+     *
+     * @param reference the reference the link points to
+     * @param label the already parsed label content
+     * @param linkParameters the parameters of the link
+     * @param freeStanding if the link is freestanding
+     */
+    public XWikiWikiReference(ResourceReference reference, XDOM label, WikiParameters linkParameters,
```
tmortagne, line 52: Even if it's not public API it's still good to have a `@since` in new reusable stuff.

tmortagne (approved):

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiWikiReference.java
michitux: Yes, I somehow forgot that, thank you. I've also added the `@since` to the new `getLabelXDOM` method.

PR #232, opened 2022-11-22
XRENDERING-690: Update TOC macro DOM root element to make it easier to style and to navigate

slauriere:
- Add CSS class to the tree generated by the TocTreeBuilder


slauriere:
This PR is a follow up to https://github.com/xwiki/xwiki-platform/pull/1869: the related issue was moved from `xwiki-platform` to `xwiki-rendering`.

@mflorea and other committers: what do you think of the CSS class name used: `wikitoc`? The naming scheme is supposed to be aligned with other class names used in `xwiki-rendering` such as `wikilink`, `wikicreatelink`.

michitux:
I think `wikitoc` is okay.

Note that this change causes test failures (see the LGTM analysis), you need to change the failing tests as they don't expect the new class attribute. xwiki-rendering tests are quite fast (iirc something like 5 minutes on my laptop) so it makes sense to run them locally before pushing.

manuelleduc (approved):

mflorea (approved):

PR #236, opened 2022-12-05
XRENDERING-693: XHTML printer does not protect against closing HTML macro in raw output

michitux:
Jira issue: https://jira.xwiki.org/browse/XWIKI-20327

I'm opening this PR primarily to make sure we agree that this escaping should be introduced here as this has some potential for breakage when the string `{{/html}}` is contained, e.g., in some JSON code that is inside an HTML macro. xwiki/xwiki-commons#329 and xwiki/xwiki-commons#323 can be understood as efforts to reduce this breakage when using XWiki APIs.

If we agree, I'll open a separate rendering issue for this change, the current issue is just the motivation for this change.


mflorea:
Should ``XHTMLWikiPrinter`` be aware of a specific wiki syntax such as XWiki 2.1? Other wiki syntaxes may use a different way to represent macro calls (i.e. using something else than ``{{...}}``, say ``%%html%%``).

PR #245, opened 2023-02-01
XRENDERING-703: Add attributes to XDOM blocks

michitux:
* Add methods for setting and getting attributes.
* Add an implementation in AbstractBlock.
* Add tests.

What I'm wondering is if there are more elegant ways to handle attributes of different types, but I think at least with generics we're out of luck.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-703


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/Block.java
```diff
@@ -280,6 +281,62 @@ public interface Block extends Cloneable
      */
     void setParameters(Map<String, String> parameters);
 
+    /**
+     * @return all attributes
+     * @since 14.10.5
+     * @since 15.1RC1
+     */
+    @Unstable
+    default Map<String, Object> getAttributes()
```
manuelleduc, line 290: wdyt of not making those methods as `default` as they don't provide anything meaningful by default?

michitux: This would be an API breakage and I wanted to avoid that, in particular as I need this change on the LTS branch for an important bug fix. Basically, what I implement by default is the behavior of an empty read-only attribute map.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/AbstractBlock.java
```diff
@@ -399,6 +404,40 @@ public void setParameters(Map<String, String> parameters)
         }
     }
 
+    @Override
+    public Map<String, Object> getAttributes()
+    {
+        return this.attributes == null ? Collections.emptyMap()
```
manuelleduc, line 410: I'm wondering is there is some meaningful differences with `Map.of()/Map.copyOf`.

michitux: I don't think so, to be honest I've just copied the code for parameters and change the names.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/Block.java
```diff
@@ -280,6 +281,62 @@ public interface Block extends Cloneable
...
+     *
+     * @param name the attribute's name
+     * @param value the attribute's value
+     * @since 14.10.5
+     * @since 15.1RC1
+     */
+    @Unstable
+    default void setAttribute(String name, Object value)
+    {
+        throw new UnsupportedOperationException();
```
tmortagne, line 323: There is not much point in implementing this method to indicate that it's not implemented. Would be much more clear to have it listed as breakage.

michitux: Okay, I've opened a [vote](https://forum.xwiki.org/t/api-breakage-on-block-to-add-attributes/11830) for the API breakage.

PR #246, opened 2023-02-01
XRENDERING-688: Improve footnote rendering

michitux:
* Parse and execute the footnote's content directly instead of from the putFootnotes macro.
* Use the id generator to generate unique ids.
* Adapt the existing test to expect nested footnote support.
* Add new tests to test some more corner cases.

Note that this requires #245 to compile.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-688


xwiki-rendering-macros/xwiki-rendering-macro-footnotes/src/main/java/org/xwiki/rendering/internal/macro/footnote/PutFootnotesMacro.java
```diff
@@ -146,26 +161,44 @@ public List<Block> execute(FootnoteMacroParameters parameters, String content, M
...
     {
-        String content = footnoteMacro.getContent();
-        if (StringUtils.isBlank(content)) {
-            content = " ";
+        @SuppressWarnings("unchecked")
+        List<Block> content = (List<Block>) footnoteMacro.getAttribute(FOOTNOTE_CONTENT_ATTRIBUTE);
+        if (content == null) {
+            // Copy the list of children as the code will later clear the original children list.
+            content = List.copyOf(footnoteMacro.getChildren());
+            footnoteMacro.setAttribute(FOOTNOTE_CONTENT_ATTRIBUTE, content);
```
michitux, line 173: This won't be cloned during a `clone()`-call which is probably bad. For nested footnote support we also kind of need that the nested footnote macro marker inside the attribute is the same object as the one we have as child of the `putFootnotes` macro as the latter is the one that is discovered by the code. I'm wondering now if it really makes sense to store these blocks in an attribute or if we shouldn't just get them from the `putFootnotes` macro. We could still store the ids in attributes such that we can more easily associate the recovered footnote content with the original footnote marker.

PR #248, opened 2023-02-03
XRENDERING-704: Provide a mechanism to customize the toc entries modularily

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-704

Note a PR will be created for xwiki-platform-export-pdf-api on xwiki-platform

**TODO**
- [x] Missing tests

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/AbstractTocMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/AbstractTocMacro.java
@@ -23,9 +23,9 @@ import java.util.List;
 import java.util.Set;
 
 import javax.inject.Inject;
-import javax.inject.Named;
 import javax.inject.Provider;
 
+import org.xwiki.component.manager.ComponentLookupException;
 import org.xwiki.component.phase.InitializationException;
 import org.xwiki.rendering.block.Block;
 import org.xwiki.rendering.block.XDOM;
@@ -34,8 +34,7 @@ import org.xwiki.rendering.listener.reference.ResourceReference;
 import org.xwiki.rendering.macro.AbstractMacro;
 import org.xwiki.rendering.macro.MacroExecutionException;
 import org.xwiki.rendering.macro.toc.TocMacroParameters;
-import org.xwiki.rendering.parser.Parser;
-import org.xwiki.rendering.renderer.reference.link.LinkLabelGenerator;
+import org.xwiki.rendering.macro.toc.TocTreeBuilderFactory;
 import org.xwiki.rendering.transformation.MacroTransformationContext;
 import org.xwiki.rendering.wiki.WikiModel;
 import org.xwiki.rendering.wiki.WikiModelException;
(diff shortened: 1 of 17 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocEntryDecorator.java
```diff
@@ -0,0 +1,50 @@
...
+package org.xwiki.rendering.macro.toc;
+
+import java.util.List;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.rendering.block.Block;
+import org.xwiki.rendering.block.HeaderBlock;
+
+/**
+ * Decorator taking an entry and manipulating the list of blocks used to render the header.
```
vmassol, line 29: You cannot document a decorator saying it's a decorator ;) What does it do?

A decorator seems to adds methods/features, see https://refactoring.guru/design-patterns/decorator. Is it the right name?

Question: why do you need to generate several blocks? Also why are generated blocks not `HeaderBlock`s?

Also, I think it would help if the javadoc was listing some examples of decorators, to understand better what it does. Examples go a long way to explain stuff.

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocTreeBuilderFactory.java
```diff
@@ -0,0 +1,53 @@
...
+        return build(null);
+    }
+
+    /**
+     * @param resolverHint the hint to use to resolve the {@link TocEntriesResolver}
+     * @return a toc tree builder using the a {@link TocEntriesResolver} with hint {@code resolverHint}
+     * @throws ComponentLookupException in case of error when accessing the {@link TocEntriesResolver} or
+     *     {@link TocEntryDecorator} instances
+     */
+    TocTreeBuilder build(String resolverHint) throws ComponentLookupException;
```
vmassol, line 52: hmm if a hint can be passed it means that what is returned is a component. In this case why do you need a factory component?

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/AbstractTocMacro.java
```diff
@@ -91,7 +80,11 @@ public void initialize() throws InitializationException
     {
         super.initialize();
 
-        this.tocTreeBuilder = new TocTreeBuilder(new TocBlockFilter(this.plainTextParser, this.linkLabelGenerator));
+        try {
+            this.tocTreeBuilder = this.tocTreeBuilderFactory.build();
+        } catch (ComponentLookupException e) {
+            throw new InitializationException(String.format("Failed to initialize [%s]", TocTreeBuilder.class), e);
```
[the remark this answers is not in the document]
manuelleduc: I'm not sure to follow. This code is in an `initialization` method and we don't have much more context to show at this point (at least no page or parameters).

xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/macro/toc/TocEntryDecorator.java
manuelleduc: > You cannot document a decorator saying it's a decorator ;) What does it do?
>
> A decorator seems to adds methods/features, see https://refactoring.guru/design-patterns/decorator. Is it the right name?

Indeed that's not a good idea to reuse the name of a know pattern.

WDYT of `Enhancer` or `Extension`?

> Question: why do you need to generate several blocks? Also why are generated blocks not HeaderBlocks?

The list of blocks is the content of the `LinkBlock` used for a ToC entry (linking to the corresponding header).

> Also, I think it would help if the javadoc was listing some examples of decorators, to understand better what it does. Examples go a long way to explain stuff.

+1

PR #251, opened 2023-02-17
minor: make private class final with default constructor

Kevin222004:
minor: make private class final with default constructor

This is part of https://github.com/checkstyle/checkstyle/pull/12737

According to the https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.8.9 The default constructor has the same access modifier as the class.
and according to the check https://checkstyle.org/config_design.html#FinalClass a class that has only private constructors and has no descendant classes is declared as final.

### DIFF
--- a/xwiki-rendering-api/src/test/java/org/xwiki/rendering/listener/chaining/AbstractChainingListenerTest.java
+++ b/xwiki-rendering-api/src/test/java/org/xwiki/rendering/listener/chaining/AbstractChainingListenerTest.java
@@ -44,7 +44,7 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
  */
 class AbstractChainingListenerTest
 {
-    private static class AbstractChild extends AbstractChainingPrintRenderer
+    private static final class AbstractChild extends AbstractChainingPrintRenderer
     {
         boolean called;
 
@@ -55,7 +55,7 @@ class AbstractChainingListenerTest
         }
     }
 
-    private static class Child extends AbstractChainingListener
+    private static final class Child extends AbstractChainingListener
     {
         boolean called;
 
@@ -66,12 +66,12 @@ class AbstractChainingListenerTest
         }
     }
 
-    private static class Child2 extends AbstractChild
+    private static final class Child2 extends AbstractChild
     {
 
     }
 
-    private static class ChildWithBothBeginMethods extends AbstractChild
+    private static final class ChildWithBothBeginMethods extends AbstractChild
     {
         boolean calledWithoutParameter;
         boolean calledWithParameter;
@@ -93,7 +93,7 @@ class AbstractChainingListenerTest
         }
     }
 
-    private static class Child3 extends AbstractChainingListener
+    private static final class Child3 extends AbstractChainingListener
     {
         boolean called;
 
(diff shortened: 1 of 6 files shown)


xwiki-rendering-api/src/test/java/org/xwiki/rendering/listener/chaining/AbstractChainingListenerTest.java
```diff
@@ -44,7 +44,7 @@
  */
 class AbstractChainingListenerTest
 {
-    private static class AbstractChild extends AbstractChainingPrintRenderer
+    private static final class AbstractChild extends AbstractChainingPrintRenderer
```
michitux, line 47: This class cannot be final because `Child2` inherits from it. This change doesn't compile.

tmortagne, line 47: This class cannot be final as it's extended by other child classes, I did not notice when I merged it (I had hope that you tried to build it actually). Would be great if you could fix that @Kevin222004

xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/impl/WikiScannerContext.java
```diff
@@ -41,7 +41,7 @@ public class WikiScannerContext implements IWikiScannerContext
 
     protected final Deque<IWikiScannerContext> fStack = new ArrayDeque<IWikiScannerContext>();
 
-    private class DefaultSectionListener extends SectionListener<WikiParameters>
+    private final class DefaultSectionListener extends SectionListener<WikiParameters>
```
tmortagne, line 44: This change also cause a build failure, reverting.

xwiki-rendering-api/src/test/java/org/xwiki/rendering/listener/chaining/AbstractChainingListenerTest.java
Kevin222004: no issues let me study i have followed only regression report. and not build code in my local but give me time will study and fix it

Kevin222004: ohh.. thats was a huge mistake from my side. i have to update the code of checkstyle to fix this

Kevin222004:
@tmortagne please tell me if the repo of 
xwiki-platform
xwiki-common is causing

xwiki-rendering-api/src/test/java/org/xwiki/rendering/listener/chaining/AbstractChainingListenerTest.java
Kevin222004: @michitux that was my mistake

tmortagne:
> @tmortagne please tell me if the repo of
> xwiki-platform
> xwiki-common is causing

Looks like you were luckier on those :)

Kevin222004:
@tmortagne thanks again for pointing i have fixed and pushed you won't effect while updating new version of checkstyle :)

PR #253, opened 2023-02-20
XRENDERING-707: The toc macro use the name of the page for empty headers entries

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-707

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocBlockFilter.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocBlockFilter.java
@@ -54,8 +54,8 @@ public class TocBlockFilter extends PlainTextBlockFilter
     {
         List<Block> children = headerBlock.clone(this).getChildren();
         if (children.isEmpty()) {
-            // In case of empty header, explicitly assign an empty label. Otherwise, the toc entry is generated with 
-            // the name of the page.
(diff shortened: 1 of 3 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/TocBlockFilter.java
```diff
@@ -51,6 +52,12 @@ public TocBlockFilter(Parser plainTextParser, LinkLabelGenerator linkLabelGenera
      */
     public List<Block> generateLabel(HeaderBlock headerBlock)
     {
-        return headerBlock.clone(this).getChildren();
+        List<Block> children = headerBlock.clone(this).getChildren();
+        if (children.isEmpty()) {
+            // In case of empty header, explicitly assign an empty label. Otherwise, the toc entry is generated with 
+            // the name of the page.
+            children = List.of(new WordBlock(""));
```
michitux, line 59: This seems like a bad hack imho. I think a `Wordblock` should be non-empty. Also, with this hack, the generated links won't be clickable in any way. I see two options:

* Use a space (i.e., `SpaceBlock`). While not really nice, it would at least ensure that you can actually click on the link I think.
* Omit empty headings in the table of contents.

manuelleduc: Thanks, https://forum.xwiki.org/t/behavior-of-the-table-of-content-in-case-of-empty-header/11977 created to discuss this.

PR #256, opened 2023-03-06
XWIKI-19383: Display a title on createlink

Sereza5:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-19383

# Changes

## Description

* Added a `WantedLinkTitleGenerator` role, resolved by a hint equal to the scheme of the reference,
  so that the title of a link to a missing resource can be adapted to the kind of resource.
* Added a default implementation naming the resource after the reference itself, used when the
  rendering runs on its own.
* Set that title on the anchor of a wanted link, so it reaches the accessible description of the
  link and not only its hover tooltip.
* Stopped the XHTML parser from turning that generated title into a link parameter.

## Clarifications

* **This PR has to be merged before xwiki/xwiki-platform#2104**, which provides the translated
  implementations and only compiles once this is in.
* The title goes on the `<a>` rather than on the wrapping `<span class="wikicreatelink">` because
  the accessible name and description of a link are computed from the link element alone —
  [accname 1.2 step 2.I](https://www.w3.org/TR/accname-1.2/#step2) does not walk up to ancestors.
  On the span the tooltip reached mouse users but no screen reader, on an issue labelled `wcag`.
  The `?` marker is `.wikicreatelink a:after`, inside the anchor, so hover loses nothing.
* Anchor attributes are semantic on the way back in: the XHTML parser turns each one into a link
  parameter, so without a counterpart in the parser an editor round trip would have written
  `||title="Create page: X"` into the user's wiki source. `XWikiReferenceTagHandler` now drops a
  `title` when the anchor's parent is a `wikicreatelink` span, following the same
  `removeMeaningfulParameters` shape already used for generated ids and freestanding markers.
  Consequence worth knowing: a hand-authored `title` on a link to a missing page is stripped too.
* One design point is still open with @vmassol — whether the by-scheme lookup is worth keeping at
  all: https://github.com/xwiki/xwiki-rendering/pull/256#discussion_r1152126160

# Screenshots & Video

No media: the change adds a tooltip and an accessible description, with no visual change to the
page. The rendered output is shown in the Executed Tests section instead.

# Executed Tests

```
mvn clean install -Plegacy,quality -pl xwiki-rendering-api,\
xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml,xwiki-rendering-integration-tests
```

`BUILD SUCCESS` on the three modules: 426 tests in `xwiki-rendering-api`, 185 in
`xwiki-rendering-syntax-xhtml`, 765 in `xwiki-rendering-integration-tests`, 0 Checkstyle
violations, all coverage checks met, Revapi clean.

New unit tests: `DefaultWantedLinkTitleGeneratorTest`, and `DocumentXHTMLLinkTypeRendererTest`
covering the three branches — an existing document gets no title, a wanted link falls back to the
default generator, and a scheme-hinted generator wins over it.

Six `wiki/link/*.test` fixtures are updated and now assert two things rather than one, the rendered
output *and* the round trip back through the parser. That second assertion is what caught the link
parameter problem described above; without it the change looked green.

Also ran the test @vmassol asked about,
`XWikiDocumentTest` in `xwiki-platform-legacy-oldcore`: 7 tests, 0 failures against these changes.
It is unaffected because its wanted-link assertion now expects `wikiexternallink` — the test has no
`WikiModel`, so it never takes the internal-link path this change touches.

Unrelated caveat: a commons SNAPSHOT from 2026-09-06 pulls surefire 3.6.0, which fails every module
with `ClassNotFoundException: org.xwiki.test.CaptureConsoleRunListener` before any test runs, on a
clean master checkout too. The runs above used `-Dsurefire.version=3.5.6`.

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * None — this adds a new API, so it targets master only.


Sereza7:
**PR Changes:**
* Solved the mismatch in hints for the URITitleGenerator implementation

**NOTE:** This commit needs to be merged with/before https://github.com/xwiki/xwiki-platform/pull/2104

Default behavior is to just use the name of the page to create as the title.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURITitleGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.rendering.listener.reference.ResourceReference;
+import org.xwiki.rendering.renderer.reference.link.URITitleGenerator;
+
+/**
+ * Generate link titles for ATTACH URIs.
+ *
+ * @version $Id$
+ * @since 15.2-RC1
```
surli, line 33: The format is not good for RC  version, it should be `15.2RC1`. Note that this format is only used in Java files, for Maven and other places we use `15.2-rc-1` format.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURITitleGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@Component
+@Named("doc")
+@Singleton
+public class DefaultPageAttachmentURITitleGenerator implements URITitleGenerator
+{
+    @Override
+    public String generateCreateTitle(ResourceReference reference)
+    {
+        return "Create page: " + reference.getReference();
```
surli, line 43: There should be a translation used here no?

xwiki-rendering-api/src/main/java/org/xwiki/rendering/renderer/reference/link/URITitleGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Generate Resource Reference titles for URIs. For example an implementation for MAILTO URIs would remove the scheme
+ * part and the query string part.
+ *
+ * @version $Id$
+ * @since 15.2RC1
+ */
+@Role
+public interface URITitleGenerator
```
surli, line 33: So this is an API, it should be marked  `@Unstable`.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURITitleGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+import org.xwiki.rendering.renderer.reference.link.URITitleGenerator;
+
+/**
+ * Generate link titles for ATTACH URIs.
+ *
+ * @version $Id$
+ * @since 15.2-RC1
+ */
+@Component
+@Named("doc")
```
surli, line 36: AFAICS this is the default component, so the one to be used "by default", so I don't think it should have a `@Named`. 
Now if it's not the default component, then you should probably rename the class.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/AbstractXHTMLLinkTypeRenderer.java
```diff
@@ -136,6 +142,31 @@ protected String computeLabel(ResourceReference reference)
...
+        // If not found then use the full reference as the label.
+        // If there's no scheme separator then use the full reference as the title. Note that this can happen
+        // when we're not in wiki mode (since all links are considered URIs when not in wiki mode).
+        String title;
+        try {
+            URITitleGenerator uriTitleGenerator =
+                this.componentManager.getInstance(URITitleGenerator.class, reference.getType().getScheme());
+            title = uriTitleGenerator.generateCreateTitle(reference);
+        } catch (ComponentLookupException e) {
+            e.printStackTrace();
```
surli, line 164: We never handle exceptions with a `printStackTrace`: you should instead log an understandable error message, using something like: 
```
this.logger.error("Error while loading component for generating URI title: [{}]", ExceptionsUtils.getRootCauseMessage(e));
this.logger.debug("Full stack trace: ", e);
```
The logger debug is not necessarily needed.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/AbstractXHTMLLinkTypeRenderer.java
```diff
@@ -136,6 +142,31 @@ protected String computeLabel(ResourceReference reference)
...
+    protected String computeCreateTitle(ResourceReference reference)
+    {
+        // Look for a component implementing URITitleGenerator with a role hint matching the link scheme.
+        // If not found then use the full reference as the label.
+        // If there's no scheme separator then use the full reference as the title. Note that this can happen
+        // when we're not in wiki mode (since all links are considered URIs when not in wiki mode).
+        String title;
+        try {
+            URITitleGenerator uriTitleGenerator =
+                this.componentManager.getInstance(URITitleGenerator.class, reference.getType().getScheme());
```
surli, line 161: you should probably handle that differently,  by first checking if the component exists with `hasComponent` and then load it only if it exists. It would avoid getting the exception for nothing. Especially if you have a default implem you could fallback in, so something like: 

```  
// in the field of the class
@Inject
private URITitleGenerator defaultTitleGenerator;

// in your method:
URITtitleGenerator titleGenerator = this.defaultTitleGenerator;
if (this.componentManager.hasComponent(URITitleGenerator.class, reference.getType().getScheme())) {
  try {
  titleGenerator = this.componentManager.getInstance(URITitleGenerator.class, reference.getType().getScheme());
  } catch (Exception e) { // handle the excepton }
}
// create the title
titleGenerator.generateCreateTitle(reference)
```

xwiki-rendering-api/src/main/java/org/xwiki/rendering/renderer/reference/link/URITitleGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.rendering.renderer.reference.link;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.rendering.listener.reference.ResourceReference;
+
+/**
+ * Generate Resource Reference titles for URIs. For example an implementation for MAILTO URIs would remove the scheme
+ * part and the query string part.
```
surli, line 27: So you indicated below: 
> // Look for a component implementing URITitleGenerator with a role hint matching the link scheme.

This should be specified in this documentation here, as it's how the implementation should be named.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURITitleGenerator.java
Sereza7: This is the default behavior, with the rendering module on its own. AFAIK there is no access to translations here. 
See [the full implementation in xwiki-platform](https://github.com/xwiki/xwiki-platform/blob/79fc276ba085daaa56c3efad5aca94265b9ca8b8/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiPageAttachmentURITitleGenerator.java#L60).

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURITitleGenerator.java
Sereza7: This is the default component used when xwiki-rendering runs on its own. This component needs to have an hint that corresponds to the kind of object I want to process. It's not the default on the resourceType axis.
I'm fixing the names of both the classes to make it clearer that the `resourceType` is `document` :+1: 
I copied the pattern used [here](https://github.com/xwiki/xwiki-rendering/blob/22eac2b1d38fbe259fa72e45df30f78eab553325/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/parser/reference/type/PageAttachmentResourceReferenceTypeParser.java#L36).

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @version $Id$
+ * @since 15.2RC1
+ */
+@Component(hints = {"doc", "page"})
+@Singleton
+public class DefaultDocumentURITitleGenerator implements URITitleGenerator
+{
+    private static final String DEFAULT_LABEL = "Create page: %s";
+    @Override
```
manuelleduc, line 40: missing line break ;)

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+ * Generate link titles for DOC URIs.
+ *
+ * @version $Id$
+ * @since 15.2RC1
+ */
+@Component(hints = {"doc", "page"})
+@Singleton
+public class DefaultDocumentURITitleGenerator implements URITitleGenerator
+{
+    private static final String DEFAULT_LABEL = "Create page: %s";
```
manuelleduc, line 39: Shouldn't that sentence be translated?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/AbstractXHTMLLinkTypeRenderer.java
```diff
@@ -46,6 +46,11 @@ public abstract class AbstractXHTMLLinkTypeRenderer implements XHTMLLinkTypeRend
      * The XHTML element <code>class</code> parameter.
      */
     protected static final String CLASS = "class";
+    /**
+     * The XHTML element <code>title</code> parameter.
```
manuelleduc, line 50:
```suggestion
     * The XHTML element {@code title} parameter.
```

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/AbstractXHTMLLinkTypeRenderer.java
```diff
@@ -46,6 +46,11 @@ public abstract class AbstractXHTMLLinkTypeRenderer implements XHTMLLinkTypeRend
      * The XHTML element <code>class</code> parameter.
      */
     protected static final String CLASS = "class";
+    /**
```
manuelleduc, line 49: missing line break

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultDocumentURITitleGenerator.java
Sereza7: https://github.com/xwiki/xwiki-rendering/pull/256#discussion_r1141852958

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/AbstractXHTMLLinkTypeRenderer.java
```diff
@@ -221,7 +226,10 @@ public void endLink(ResourceReference reference, boolean freestanding, Map<Strin
     {
         // If there was no link content then generate it based on the passed reference
         if (!hasLabel()) {
-            getXHTMLWikiPrinter().printXMLStartElement(SPAN, new String[][] { { CLASS, "wikigeneratedlinkcontent" } });
+            getXHTMLWikiPrinter().printXMLStartElement(SPAN, new String[][]
+            {
+                { CLASS, "wikigeneratedlinkcontent" }
```
manuelleduc, line 231: Is it possible to already has classes defined at this point?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -51,6 +55,8 @@ public class DocumentXHTMLLinkTypeRenderer extends AbstractXHTMLLinkTypeRenderer
      */
     private static final String WIKILINK = "wikilink";
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractXHTMLLinkTypeRenderer.class);
```
manuelleduc, line 58: Injecting  the logger is preferable since you are in a component.
```suggestion
    @Inject
    private Logger logger;
```

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -90,6 +103,31 @@ protected String computeLabel(ResourceReference reference)
...
     }
 
+    /**
+     * Implementation for computing a document link title when no title has been specified.
+     * Looks for a component implementing URITitleGenerator with a role hint matching the reference scheme.
+     * @param reference the reference of the link for which to compute the label
+     * @return the computed title
+     * @since 15.2RC1
+     */
+    private String computeCreateTitle(ResourceReference reference)
```
manuelleduc, line 113: I think it is interesting to move the resolution of the `URITitleGenerator` resolution in a separate method, and to keep only the computation of the create title in this method.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURITitleGenerator.java
manuelleduc: In this case, we should discuss a way to allow xwiki platform to inject the a translation here. Having some text in an unknown language is not going to help accessibility for non-English speaking users.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -90,6 +103,31 @@ protected String computeLabel(ResourceReference reference)
...
+        URITitleGenerator titleGenerator = this.defaultTitleGenerator;
+        String title;
+        if (this.componentManager.hasComponent(URITitleGenerator.class, reference.getType().getScheme())) {
+            try {
+                titleGenerator = this.componentManager.getInstance(URITitleGenerator.class,
+                 reference.getType().getScheme());
+            } catch (Exception e) {
+                LOGGER.error("Error while loading component for generating URI title: [{}]",
+                    ExceptionUtils.getRootCauseMessage(e));
+                LOGGER.debug("Full stack trace: ", e);
```
manuelleduc, line 124: See https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices for our rules regarding which logging level to use. I'm not sure failing to resolve a title generator is:

> an important problem that compromises the stability of the XWiki instance or that prevent an important system from working and that should not have happened

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/AbstractXHTMLLinkTypeRenderer.java
Sereza7: _Note:_ the changes here are purely quality of code, I found it hard to read out so I separated it on a few lines. This part was particularly interesting to me because I wanted to replicate a behavior similar to the call to computeLabel on line 233. 

The class defined here is the one we want for the label of the link, which is a new child element of the link. We print this label completely on these few lines.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURITitleGenerator.java
Sereza7: This behavior is overwritten by the xwiki platform class I linked above.
I followed the pattern used so far when we needed translations in xwiki-rendering.
See Michael's comment on the [jira issue](https://jira.xwiki.org/browse/XWIKI-19383).

Other similar uses of this pattern can be found [here](https://github.com/xwiki/xwiki-rendering/blob/22eac2b1d38fbe259fa72e45df30f78eab553325/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultPageAttachmentURILabelGenerator.java#L44) and [here](https://github.com/xwiki/xwiki-rendering/blob/bda9e53453ddabbe2d58490c313707f2c7a11dde/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DataURILabelGenerator.java#L44).
I thought it would be better to have, in all cases, a readable english title instead of just a value that could hold anything (possibly complete gibberish). We could remove this default formatting to just return reference.getReference() but I don't think it'd be best.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
Sereza7: Changing it to a warning :+1:

Sereza7:
**PR Changes:**
* Refactored name 
* Fixed hint to match DocumentXHTMLLinkTypeRenderer.
* Moved behavior from the AbstractXHTMLLinkTypeRenderer to DocumentXHTMLLinkTypeRenderer
* Marked the API as unstable
* Fixed code style
* Improved comments

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -62,6 +68,12 @@ public class DocumentXHTMLLinkTypeRenderer extends AbstractXHTMLLinkTypeRenderer
     @Inject
     private LinkLabelGenerator linkLabelGenerator;
 
+    /**
+     * Used to generate a link title.
+     */
+    @Inject
+    private URITitleGenerator defaultTitleGenerator;
```
vmassol, line 75: This is not going to work since there's no default implementation.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -90,6 +102,35 @@ protected String computeLabel(ResourceReference reference)
...
+    {
+        URITitleGenerator titleGenerator = this.defaultTitleGenerator;
+        if (this.componentManager.hasComponent(URITitleGenerator.class, reference.getType().getScheme())) {
+            try {
+                titleGenerator = this.componentManager.getInstance(URITitleGenerator.class,
+                    reference.getType().getScheme());
+            } catch (Exception e) {
+                logger.warn("Error while loading component for generating URI title: [{}]",
+                    ExceptionUtils.getRootCauseMessage(e));
+                logger.debug("Full stack trace: ", e);
```
vmassol, line 115: I would write:

```
String message = "Error while loading component for generating URI title";
if (logger.isDebugEnabled()) {
    logger.debug(message, e);
} else {
    logger.warn(String.format("%s: [{}]", message), ExceptionUtils.getRootCauseMessage(e));
}
```

vmassol: In addition you should also log the reference and explain which component we're talking about.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/renderer/reference/link/URITitleGenerator.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Generate Resource Reference titles for URIs.
+ * The implementations should be named according to the kind of reference they process.
+ *
+ * @version $Id$
+ * @since 15.2RC1
+ */
+@Role
+@Unstable
+public interface URITitleGenerator
```
vmassol, line 35: I don't understand why it's a URI. I'd name this: `ResourceReferenceTitleGenerator`.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+import org.xwiki.rendering.listener.reference.ResourceReference;
+import org.xwiki.rendering.renderer.reference.link.URITitleGenerator;
+
+/**
+ * Generate link titles for DOC URIs.
+ *
+ * @version $Id$
+ * @since 15.2RC1
+ */
+@Component(hints = {"doc", "page"})
```
vmassol, line 34: Error here, should not contain any hint.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/renderer/reference/link/URITitleGenerator.java
```diff
@@ -0,0 +1,42 @@
...
+import org.xwiki.component.annotation.Role;
+import org.xwiki.rendering.listener.reference.ResourceReference;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Generate Resource Reference titles for URIs.
+ * The implementations should be named according to the kind of reference they process.
+ *
+ * @version $Id$
+ * @since 15.2RC1
```
vmassol, line 31: Needs to be updated to 15.3RC1 now ;)

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/AbstractXHTMLLinkTypeRenderer.java
```diff
@@ -221,7 +227,10 @@ public void endLink(ResourceReference reference, boolean freestanding, Map<Strin
...
```
vmassol, line 232: I wouldn't change the formatting (unless there's a good reason)

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/renderer/DefaultDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+package org.xwiki.rendering.internal.renderer;
+
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.rendering.listener.reference.ResourceReference;
+import org.xwiki.rendering.renderer.reference.link.URITitleGenerator;
+
+/**
+ * Generate link titles for DOC URIs.
```
vmassol, line 29: Improve the javadoc and explain what it does (use the reference as the title) and explain that it's supposed to be overridden to provide better titles.

vmassol:
Missing some test (which explains why the code is buggy - no default implementation - but it was not noticed).

vmassol:
Have you tested that https://github.com/xwiki/xwiki-platform/blob/9330fea36561122e2a7ce38bec6fb2c6ccde48a6/xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/test/java/com/xpn/xwiki/doc/XWikiDocumentTest.java#L280 passes?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -90,6 +102,35 @@ protected String computeLabel(ResourceReference reference)
         return this.linkLabelGenerator.generate(reference);
     }
 
+    private URITitleGenerator getTitleGenerator(ResourceReference reference)
+    {
+        URITitleGenerator titleGenerator = this.defaultTitleGenerator;
+        if (this.componentManager.hasComponent(URITitleGenerator.class, reference.getType().getScheme())) {
```
vmassol, line 108: I don't see why need this as the implementation is fixed and doesn't depend on the resource reference type.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -90,6 +102,37 @@ protected String computeLabel(ResourceReference reference)
...
     }
 
+    private WantedLinkTitleGenerator getTitleGenerator(ResourceReference reference)
+    {
+        WantedLinkTitleGenerator titleGenerator = this.defaultTitleGenerator;
+        try {
+            titleGenerator = this.componentManager.getInstance(WantedLinkTitleGenerator.class,
+                reference.getType().getScheme());
+        } catch (Exception e) {
+            String message = "Could not find a [WantedLinkTitleGenerator] component to generate the wanted "
```
vmassol, line 112: you should use `WantedLinkTitleGenerator.class.getName()` instead of hardcoding it IMO.

Sereza5:
Rebased onto current master and reworked. The PR description was the state at open time; this covers
what changed since.

**The title moved from the span to the anchor.** On `<span class="wikicreatelink">` it reached mouse
users but no screen reader: [accname 1.2 step 2.I](https://www.w3.org/TR/accname-1.2/#step2) uses the
tooltip attribute of the *current node* and never walks up to ancestors. The `?` marker is
`.wikicreatelink a:after`, i.e. inside the anchor, so hover loses nothing.

**That needed a parser change — the part worth reviewing.** Anchor attributes are semantic on the way
back in: the parser turns each into a link parameter, so an editor round trip would have written
`||title="Create page: X"` into the user's wiki source. `XWikiReferenceTagHandler` now drops a
`title` when the anchor's parent is a `wikicreatelink` span, following the existing
`removeMeaningfulParameters` shape. Two paths needed it — `end()` for plain `<a href>`, and `begin()`
where `LINK_PARAMETERS` is set for the annotated path that the editor actually round-trips.

Trade-off to weigh: a hand-authored `title` on a link to a missing page is now stripped too. Say so
if you would rather distinguish authored from generated.

Tests added for both generators and for the renderer; the `wiki/link/*.test` fixtures now assert the
rendered output *and* the round trip. `@since` tags are `18.8.0RC1`, and the unrelated
`printXMLStartElement` reformatting is reverted.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/renderer/reference/link/URITitleGenerator.java
Sereza5: Agreed, fixed in 9e82f22b1 👍 The interface Javadoc now states that the hint must equal
`ResourceType#getScheme()`, and that an implementation registered under any other hint silently falls
back to the no-hint default and is never called.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
Sereza5: Fixed. There is a no-hint `DefaultWantedLinkTitleGenerator` now, returned whenever no scheme-hinted
component is registered. Covered by `DocumentXHTMLLinkTypeRendererTest` 👍

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
Sereza5: Kept — but tell me if you still disagree and I will collapse it. This is the same decision as your
*"Not needed."* on the `@Component` hints in xwiki/xwiki-platform#2104, so I am answering both here,
since this PR is the one that has to merge first.

The implementation is not fixed once `xwiki-platform` is in the picture: it registers one generator
for `doc`/`page` and another for `space`, giving "Create page: X" versus "Create space: X". `doc` and
`page` are two syntaxes for the same object so they share one class.

A single fixed implementation would have to switch on `reference.getType()` itself, which moves the
dispatch inside rather than removing it, and stops an extension adding a title for its own scheme. It
is also the shape already used around it — `AbstractXHTMLLinkTypeRenderer.computeLabel` looks
`URILabelGenerator` up by `reference.getType().getScheme()` two methods above.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
Sereza5: Done 👍 I also reworked the statement itself in 86d536579 so every value is an SLF4J parameter — see
the PR comment, since that changes the shape you originally suggested.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/renderer/xhtml/link/DocumentXHTMLLinkTypeRenderer.java
```diff
@@ -90,6 +102,37 @@ protected String computeLabel(ResourceReference reference)
         return this.linkLabelGenerator.generate(reference);
     }
 
+    private WantedLinkTitleGenerator getTitleGenerator(ResourceReference reference)
+    {
+        WantedLinkTitleGenerator titleGenerator = this.defaultTitleGenerator;
+        try {
+            titleGenerator = this.componentManager.getInstance(WantedLinkTitleGenerator.class,
+                reference.getType().getScheme());
```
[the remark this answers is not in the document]
Sereza5: You were right, it did not fall back. Fixed and tested: `getTitleGenerator` checks `hasComponent`
first and returns the default otherwise, which also stops it warning on every wanted link in the
standalone case 👍

Sereza5:
Answer to https://github.com/xwiki/xwiki-rendering/pull/256#issuecomment-1488787892 and https://github.com/xwiki/xwiki-rendering/pull/256#issuecomment-1488793749
Both fair, and both now covered.

There are unit tests for `DefaultWantedLinkTitleGenerator` and for
`DocumentXHTMLLinkTypeRenderer`, the latter covering the fallback branch whose absence you spotted,
plus tests for the three generators on the platform side. Built with `-Plegacy,quality` on every
touched module: 0 Checkstyle violations, coverage checks met, Revapi clean.

`links6.test` and five other `wiki/link/*.test` fixtures are updated, and the whole
`xwiki-rendering-integration-tests` suite passes: 765 tests. Those fixtures now assert two things
rather than one, the rendered output *and* the round trip back through the XHTML parser, which is
what caught the parser problem described in the PR comment.

`XWikiDocumentTest` passes too — I ran it: 7 tests, 0 failures, against these changes. Worth noting
why it is unaffected, since your link was to the wanted-link assertion: that assertion is now at
line 259 and expects `wikiexternallink`, not `wikicreatelink`. The test has no `WikiModel`, so the
link never takes the internal-link path, and my change is confined to the wanted-link branch inside
`beginInternalLink`. I also grepped the rest of xwiki-platform: nothing else asserts
`wikicreatelink` markup apart from the page test I added.

____

All comments have been addressed, this PR is ready for review :)
Ongoing discussions (non-blocking IMO):
* https://github.com/xwiki/xwiki-rendering/pull/256#discussion_r1152126160

vmassol (changes requested):
@Sereza5 you cannot introduce new public APIs like this, especially when fixing issues related to some visual aspects that have nothing to do with public APIs. A public API is something very serious, since we will never be able to remove it, forever (ok, after the @Unstable 1 cycle delay, but still if you push it now withut an agreement we will likely forget to review it). So it needs to be reviewed and agreed with all devs. This is why new public apis require a proposal on the forum. We'll need to agree about at least the following:
* The naming, the parameters
* Whether it's the right way to do it (ie architecture/design) or if there's a better way

Thx

Sereza5:
Note that this has minimal impact on the visual aspect. It's more about making information explicit.

I created  a discussion on the forum at https://forum.xwiki.org/t/proposal-new-api-in-rendering-to-give-wanted-links-a-proper-text-alternative/18844 . Setting this PR as a draft until the discussion is resolved.

PR #258, opened 2023-03-27
XRENDERING-655: Add the ability to style image captions

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-655
Platform PR: https://github.com/xwiki/xwiki-platform/pull/2136
Forum proposal: https://forum.xwiki.org/t/inject-chain-listeners-from-extensions/12291

**TODO**
- Link to PR for `xwiki-platform`
- move "style" specific code in `xwiki-platform`


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki21/src/main/java/org/xwiki/rendering/internal/listener/TmpListenerProvider.java
```diff
@@ -0,0 +1,107 @@
...
+
+            super.onImage(reference, freestanding, id, parameters);
+        }
+
+        @Override
+        public void endFigure(Map<String, String> parameters)
+        {
+            Map<String, String> withImageParams = new HashMap<>(parameters);
+            if (this.additionalFigureParameters != null) {
+                withImageParams.putAll(this.additionalFigureParameters);
```
michitux, line 95: The parameters in `beginFigure` and `endFigure` need to match, parameters in `endFigure` are frequently ignored.

michitux (commented):
I think these listeners (and their providers) need to be way more specific sucht hat they can target the parsing or rendering of a specific syntax.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-wikimodel/src/main/java/org/xwiki/rendering/internal/parser/wikimodel/AbstractWikiModelParser.java
```diff
@@ -131,11 +157,25 @@ public void parse(Reader source, Listener listener, IdGenerator idGenerator) thr
...
+
+    private Listener buildListener(Listener listener)
+    {
+        // TODO: to be documented...
+        ListenerChain chain = new ListenerChain();
+        ChainingWrappingListener wrappedListener = new ChainingWrappingListener();
+        wrappedListener.setWrappedListener(listener);
+        StartChainingListener startChainingListener = new StartChainingListener();
+        startChainingListener.setListenerChain(chain);
+        chain.addListener(startChainingListener);
```
michitux, line 176: Actually, after thinking again about this, I think you don't need this, you can just take the first listener of the chain and return it. I mean this way it might be a tiny bit safer in case some listener wants to inject another listener before itself, but it also means there are always two listener even if all you need is the actual listener.

As a small optimization, you could also just pass the given listener if no listeners were registered.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki21/src/main/java/org/xwiki/rendering/internal/listener/TmpParseListenerProvider.java
```diff
@@ -0,0 +1,144 @@
...
+                        computedValue = String.format(format, existingStyle, value);
+                    } else {
+                        computedValue = value;
+                    }
+                    withImageParams.put(key, computedValue);
+                });
+                withImageParams.putAll(this.additionalFigureParameters);
+                this.additionalFigureParameters = null;
+            }
+            super.endFigure(withImageParams);
```
michitux, line 129: You need to pass these same arguments also to `beginFigure`. I think what you need to do is to start queuing when you encounter `beginFigure` and only forward the events when you've reached `endFigure`. You can achieve this by inheriting from `LookaheadChainingListener` which supports queuing, see `StackingFigureContentChainingListener` for an example.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/listener/ListenerRegistry.java
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public static final String PARSE_ACTION = "parse";
+
+    /**
+     * Render action identifier.
+     */
+    public static final String RENDER_ACTION = "render";
+
+    @Inject
+    private ComponentManager componentManager;
```
tmortagne, line 57: Shouldn't you use a "context" ComponentManager here ? Otherwise, it will only find component registered in the root namespace.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki21/src/main/java/org/xwiki/rendering/internal/listener/TmpParseListenerProvider.java
michitux: For example, the HTML renderer only prints the arguments that are passed to `beginFigure`, the arguments to `endFigure` are ignored.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/listener/ListenerRegistry.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Inject
+    private Logger logger;
+
+    /**
+     * Register a list of {@link ChainingListener} provided by {@link ListenerProvider} in the provided chain.
+     *
+     * @param listenerChain the listener chain in which new listener will be added
+     * @param action the action performed by the caller ("parse" or "render")
+     * @param syntaxHint the hint of the syntax of the action (e.g., "xwiki/2.1")
```
tmortagne, line 67: Why not pass a Syntax ?

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/ListenerProvider.java
```diff
@@ -0,0 +1,50 @@
...
+@Role
+@Unstable
+public interface ListenerProvider
+{
+    /**
+     * @param action the action performed by the listener ("render" or "parse")
+     * @param syntaxHint the hint of the syntax using for the action
+     * @return {@code true} when the listener provider can return a listener for the given action and syntaxHint
+     */
+    boolean accept(String action, String syntaxHint);
```
tmortagne, line 43: Why not pass a Syntax ?

manuelleduc: Indeed, might be cleaner. The only cons is that we now need to parse the hints.

michitux, line 43: Wouldn't it be better to either use an enum for the action or at least provide constants for the two possible values directly in `ListenerProvider`?

manuelleduc: The constants are here https://github.com/xwiki/xwiki-rendering/pull/258/files#diff-a69d93e25fa580489bfec51a5d51a405b7ad87f7f2f5b8ad0592f768a2aea9d2R49-R54
I considered the enum but I'm not sure of the pros and cons here.

michitux: Yes, I've seen that you've provided the constants in `ListenerRegistry`, but `ListenerRegistry` is internal so any extension that wants to provide a `ListenerProvider` either needs to depend on this internal class or re-define the constants. Now I'm wondering if `ListenerRegistry` shouldn't also be public API as it might be interesting to use for any extension that implements a parser or renderer.

manuelleduc: Indeed, I meant to have this constants public so either:
- moved them to `ListenerProvider`
- make `ListenerRegistry` public

xwiki-rendering-syntaxes/xwiki-rendering-syntax-wikimodel/src/main/java/org/xwiki/rendering/internal/parser/wikimodel/AbstractWikiModelParser.java
```diff
@@ -131,11 +167,32 @@ public void parse(Reader source, Listener listener, IdGenerator idGenerator) thr
...
+        String roleHint = this.descriptor.getRoleHint();
+        Syntax syntax = this.syntaxRegistry.getSyntax(roleHint).orElse(null);
+        if (syntax == null) {
```
tmortagne, line 189: It's a bit surprising to not use Optional#isEmpty since that was the kind of the point of returning an Optional in this API :)

xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/ListenerProvider.java
manuelleduc: Done

manuelleduc: Constants moved to `ListenerProvider`.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-wikimodel/src/main/java/org/xwiki/rendering/internal/parser/wikimodel/AbstractWikiModelParser.java
manuelleduc: +1, though this is now replaced by the use of `resolveSyntax` which fallback to parsing if no matching syntax is registered, and throws a `ParseException` error if everything else failed.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki21/pom.xml
```diff
@@ -31,7 +31,8 @@
   <name>XWiki Rendering - Syntax - XWiki 2.1</name>
   <description>XWiki Rendering - Syntax - XWiki 2.1</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.47</xwiki.jacoco.instructionRatio>
+    <!-- Reverse back when the provider are moved to platform. -->
```
michitux, line 34: This looks like a forgotten todo comment.

PR #259, opened 2023-04-03
XRENDERING-712: annotated xhtml Inline macro following inline parameters can be lost when rendered in xwiki syntax

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-712
Initially raised issue: https://jira.xwiki.org/browse/XWIKI-20805


tmortagne:
I would have expected to see a new unit test in xwiki-rendering-syntax-xhtml itself.

michitux:
Are you sure that in a non-generated content block there is no chance there will be generated content inside?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiCommentHandler.java
```diff
@@ -112,18 +113,12 @@ public void onComment(String content, TagStack stack)
 
     private void handleMacroCommentStart(String content, TagStack stack)
     {
-        boolean shouldIgnoreAll;
 
-        // true if we are already in a non generated content block
-        boolean inNonGeneratedContent = stack.getStackParameter(NON_GENERATED_CONTENT_STACK) != null
-            && (Boolean) stack.getStackParameter(NON_GENERATED_CONTENT_STACK);
+        boolean inNonGeneratedContent = isInNonGeneratedContent(stack);
```
surli, line 117: Looks like this is only used for `shouldIgnoreAll`, so why not instead create a method to compute `shouldIgnoreAll` value?

manuelleduc: Why not. But the first question to answer is, how to accurately compute `inNonGeneratedContent`?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiCommentHandler.java
```diff
@@ -302,4 +293,25 @@ private String computeAltAttributeValue(ResourceReference reference)
...
+                        // Stops at the first value which is associated with a macro.
+                        if (value.isMetaDataElement) {
+                            return value.nonGeneratedContent;
+                        }
+                    }
+                    return false;
+                }).orElse(false);
+
+        // If we are in a macro but not in a non generated content, we should ignore all
+        return stack.getStackParameter(MACRO_INFO) != null && !inNonGeneratedContent;
```
michitux, line 315: Doesn't every div push a new stack and thus "remove" the `MACRO_INFO` stack parameter?

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiCommentHandler.java
```diff
@@ -302,4 +293,25 @@ private String computeAltAttributeValue(ResourceReference reference)
...
+    private boolean shouldIgnoreAll(TagStack stack)
+    {
+        Optional<Iterator<Object>> stackParameterIterator =
+            Optional.ofNullable(stack.getStackParameterIterator(NON_GENERATED_CONTENT_STACK));
+        boolean inNonGeneratedContent =
+            stackParameterIterator
+                .map(iterator -> {
+                    while (iterator.hasNext()) {
+                        NonGeneratedContentStackValue value = (NonGeneratedContentStackValue) iterator.next();
+                        // Stops at the first value which is associated with a macro.
```
michitux, line 306: Why is a metadata element associated with a macro? Couldn't this also be independent of a macro?

michitux (commented):
I'm not convinced that this solution is correct (see my comments) but I haven't checked the tests.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiCommentHandler.java
manuelleduc: Does it?

michitux: See https://github.com/xwiki/xwiki-rendering/blob/045ae68924e6bc80fb2ccea17f64349817244c06/xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/impl/WikiScannerContext.java#L149 and https://github.com/xwiki/xwiki-rendering/blob/045ae68924e6bc80fb2ccea17f64349817244c06/xwiki-rendering-wikimodel/src/main/java/org/xwiki/rendering/wikimodel/impl/WikiScannerContext.java#L704

manuelleduc: Thanks. So I need to find a way to visit the stack of macro calls, and not the internal state of a single macro.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiCommentHandler.java
manuelleduc: Indeed, I just noticed that this can also be used for macro parameters.

PR #262, opened 2023-04-26
XRENDERING-716: Move the figure macro in a contrib extension

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-716

I'm not sure if we should also apply the change on `stable-14.10.x` too.


tmortagne:
> I'm not sure if we should also apply the change on stable-14.10.x too.

Would be very strange to do such a change in a LTS IMO.

vmassol:
> I'm not sure if we should also apply the change on stable-14.10.x too.

Ideally, we need to make sure the new figure macro works on 14.10.x, somehow.

manuelleduc:
> Ideally, we need to make sure the new figure macro works on 14.10.x, somehow.

The minimal version of `macro-figure` is 14.9. I'll run some additional tests but user wishing to upgrade to the contrib version on 14.10.x should be able to do so.

vmassol:
> The minimal version of macro-figure is 14.9. I'll run some additional tests but user wishing to upgrade to the contrib version on 14.10.x should be able to do so.

What happens if the figure macro is already installed, will it work?

manuelleduc:
> What happens if the figure macro is already installed in a 14.10.x instance and the new contrib figure extension is installed, will it work? Or do we need to provide instructions to uninstall it first?
Thanks to `<xwiki.extension.features>org.xwiki.rendering:xwiki-rendering-macro-figure</xwiki.extension.features>` defined in `macro-figure` + a first release version of 15.4, it should be possible for all versions >= 14.9 to install `macri-figure`.
I've tried on a 14.10.8 using the advanced search to point to my local `.m2` and the UI offers to upgrade the extensions nicely.

The part if note sure is what's going to happen when the extension will be released and installed in e.x.o.
@tmortagne will the use see to "Figure Macro" entries in the search UI?

tmortagne:
>  @tmortagne will the use see to "Figure Macro" entries in the search UI?

Depends what the user search for, but I assume searching for "figure" will get it. I'm not really sure why you ask, actually, why would users not see a compatible extension ?

manuelleduc:
> Depends what the user search for, but I assume searching for "figure" will get it. I'm not really sure why you ask, actually, why would users not see a compatible extension ?

Mainly for the ambiguity. If they see two "Figure Macro" entries, how can the decide of the right one to install?

tmortagne:
> If they see two "Figure Macro" entries

They will see two "Figure Macro" entries only if you have two "Figure Macro" entries pages on extensions.xwiki.org.

manuelleduc:
> They will see two "Figure Macro" entries only if you have two "Figure Macro" extensions/pages on extensions.xwiki.org (I don't think that's a good idea).

Ah right, forgot about that option, thanks. Looks like everything is good then!

PR #263, opened 2023-05-05
XRENDERING-719: Make the numbered property advanced

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-719

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -276,7 +276,19 @@
                  </revapi.differences>
             -->
 
-            
+            <revapi.differences>
+              <differences>
+                <item>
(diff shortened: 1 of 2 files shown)


michitux:
> Note: currently, even with the legacy module installed, the `numbered` parameter is never used, meaning that the numbered will never be displayed.

That seems very bad as it means (according to my understanding) that users will continue to see the numbered parameter as a regular parameter in the macro dialog but it won't have any effect. So from a user's point of view, this is a regression. I think we should either continue to fully support the numbered parameter (optionally hidden as advanced) or fully remove it from the UI.

PR #266, opened 2023-06-02
XRENDERING-670: Image captions cannot be used in combination with links on images

michitux:
* Accept figures where the image is nested inside a link as image figures.
* Allow XWiki syntax rendering to produce nested links to allow links in captions of images that are wrapped inside a link.
* Introduce a new XWiki20XWikiGeneratorListener to handle the transformation to figures during XWiki syntax parsing.
* Store the image label in a parameter to transfer it from the link parser to the main parser.
* Add tests.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-670

### DIFF
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/main/java/org/xwiki/rendering/internal/parser/xwiki20/wikimodel/XWiki20XWikiGeneratorListener.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/main/java/org/xwiki/rendering/internal/parser/xwiki20/wikimodel/XWiki20XWikiGeneratorListener.java
@@ -1,0 +1,189 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/main/java/org/xwiki/rendering/internal/parser/xwiki20/wikimodel/XWiki20XWikiGeneratorListener.java
```diff
@@ -0,0 +1,189 @@
...
+                    @Override
+                    public void endDocument(MetaData metadata)
+                    {
+                        // ignore.
+                    }
+                };
+                wrapper.setWrappedListener(getListener());
+                this.imageLabelParser.parse(new StringReader(label), wrapper, getIdGenerator());
+            } catch (ParseException e) {
+                // TODO what should we do here ?
```
manuelleduc, line 173: Good question. This is the case where the label of an image link fails to be parsed right?

michitux: Note that this code is not new, I've just moved it. The TODO was already there before, and we have the same TODO for link content which is where I got this from.

Ignoring the exception here basically means that the content will be lost during rendering if there is an exception. A better idea is probably to display an error block with exception information as we do in other places.

manuelleduc: ok, so an interesting improvement, but definitely a separate issue. Thanks

PR #272, opened 2023-07-10
XRENDERING-721: Content of script tags should be ignored to avoid additional content being saved

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-721

**TODO**
- [x] add for html5
- [x] also exclude `script` elements
- [x] make sure to support nested elements (not needed for the current use cases, but still safer to have it work as expected)

### DIFF
--- a/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiHeadTagHandler.java
+++ b/xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiHeadTagHandler.java
@@ -1,0 +1,62 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.rendering.internal.parser.xhtml.wikimodel;
+
+import java.util.Objects;
+
+import org.xwiki.rendering.wikimodel.xhtml.handler.TagHandler;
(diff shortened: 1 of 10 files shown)


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiHeadTagHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    protected void begin(TagContext context)
+    {
+        TagStack tagStack = context.getTagStack();
+        tagStack.pushIgnoreElementRule(new IgnoreElementRule(ignoreElementRule -> {
+            TagContext tagContext = ignoreElementRule.getTagContext();
+            // Pops itself off the stack when the head element ends.
+            if (Objects.equals(tagContext.getName(), TAG_HEAD)) {
```
manuelleduc, line 56: Check if the head is a closing element and not a nested one.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/test/resources/xhtml10/specific/head/head1.test
```diff
@@ -0,0 +1,22 @@
+.#-----------------------------------------------------
+.input|xhtml/1.0
+.#-----------------------------------------------------
```
michitux, line 3: It would be nice to have a comment here on what's tested in this test.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/test/resources/xhtml10/specific/script/script1.test
```diff
@@ -0,0 +1,22 @@
+.#-----------------------------------------------------
+.input|xhtml/1.0
+.#-----------------------------------------------------
```
michitux, line 3: Same here, a comment what's tested would be nice.

xwiki-rendering-syntaxes/xwiki-rendering-syntax-xhtml/src/main/java/org/xwiki/rendering/internal/parser/xhtml/wikimodel/XWikiIgnoredTagHandler.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.Objects;
+
+import org.xwiki.rendering.wikimodel.xhtml.handler.TagHandler;
+import org.xwiki.rendering.wikimodel.xhtml.impl.IgnoreElementRule;
+import org.xwiki.rendering.wikimodel.xhtml.impl.TagContext;
+import org.xwiki.rendering.wikimodel.xhtml.impl.TagStack;
+
+/**
+ * Handle any tag for which the content can be ignored during parsing.
```
michitux, line 30:
```suggestion
 * Handle any tag for which the content should be ignored during parsing.
```

michitux (approved):
Looks good to me, just some minor comments regarding the comments/documentation.

I'm not sure that this is the intended way ignore element rules should be used but to me it looks like it should be working this way.

PR #282, opened 2023-10-05
XRENDERING-723: Allow preparing macros

tmortagne:
issues:
* https://jira.xwiki.org/browse/XRENDERING-703
* https://jira.xwiki.org/browse/XRENDERING-723

This adds to the rendering API the concept of block preparation: inject in the block tree attributes containing anything that can be processed without any contextual information so that it can be cached and used later during the actual translation.


xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/Block.java
```diff
@@ -280,6 +281,54 @@ enum Axes
      */
     void setParameters(Map<String, String> parameters);
 
+    /**
+     * @return all attributes
+     * @since 15.9-rc-1
```
michitux, line 286: Don't we use the 15.9RC1 format here (same for other since-comments)?

xwiki-rendering-api/src/main/java/org/xwiki/rendering/transformation/Transformation.java
```diff
@@ -64,4 +64,16 @@ public interface Transformation extends Comparable<Transformation>
      * @since 2.4M1
      */
     void transform(Block block, TransformationContext context) throws TransformationException;
+
+    /**
+     * Prepare a block meant to be cached to be executed several times. The goal is to pre-execute everything that can
+     * fit in any context and generally store them in block attributes (which are nor serialized).
```
michitux, line 70:
```suggestion
     * fit in any context and generally store them in block attributes (which are not serialized).
```

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
```diff
@@ -354,4 +358,69 @@ public void setMaxRecursions(int maxRecursions)
...
+    public void prepare(Block block)
+    {
+        Syntax syntax = null;
+
+        // Find the current syntax
+        MetaDataBlock metaDataBlock =
+            block.getFirstBlock(new MetadataBlockMatcher(MetaData.SYNTAX), Axes.ANCESTOR_OR_SELF);
+        if (metaDataBlock != null) {
+            syntax = (Syntax) metaDataBlock.getMetaData().getMetaData(MetaData.SYNTAX);
+        }
```
michitux, line 372: A very similar logic exists in `org.xwiki.rendering.internal.macro.DefaultMacroContentParser#getCurrentSyntax` - wouldn't it make sense to use that method or to refactor the code to avoid the duplication?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
```diff
@@ -354,4 +358,69 @@ public void setMaxRecursions(int maxRecursions)
...
+            Macro<?> macro = knownMacros.get(macroBlock.getId());
+
+            // If not found use the macro manager
+            try {
+                if (macro == null) {
+                    macro =
+                        MacroTransformation.this.macroManager.getMacro(new MacroId(macroBlock.getId(), currentSyntax));
+
+                    // Cache the found macro for later
+                    knownMacros.put(macroBlock.getId(), macro);
```
michitux, line 404: Isn't it dangerous/incorrect to cache macros by id without syntax but to request them with syntax?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
```diff
@@ -354,4 +358,69 @@ public void setMaxRecursions(int maxRecursions)
...
+            try {
+                macro.prepare(macroBlock);
+            } catch (MacroPreparationException e) {
+                this.logger.error("Failed to prepare the macro block", e);
+            }
+        }
+
+        // Prepare the children
+        for (Block child : block.getChildren()) {
+            prepare(child, currentSyntax, knownMacros);
```
michitux, line 423: Isn't it dangerous that this recursion could hit the stack limit? Or do we already have other code where we would similarly hit the stack limit?

michitux (approved):
Looks good overall, but I think many since-comments use the wrong version format and I have some doubts regarding the macro cache and the recursion.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/transformation/Transformation.java
```diff
@@ -64,4 +64,16 @@ public interface Transformation extends Comparable<Transformation>
...
+     * 
+     * @param block the block to prepare (can be an {@link XDOM})
+     * @since 15.9-rc-1
+     */
```
michitux, line 74: Missing `@Unstable`.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: Yes I thought about it but "we already have other code where we would similarly hit the stack limit" so I did not spend the time working on a better navigation (I would have had to fix all the other places).

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: I actually partially copy/pasted the logic we have in transform but indeed you are right, it's not accurate as the syntax might change (even if I doubt we have macros associated with syntaxes currently).

To be honest I don't think that local cache really worth it compared to simply always calling MacroManager#getMacro, so I will remove it.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: We have the problem for example when rendering an XDOM because of `AbstractBlock#traverse`.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: The main point of `DefaultMacroContentParser#getCurrentSyntax` is to find the syntax in the current context while here all we want is to find the syntax of a given block, which is `#getFirstBlock`'s job (which, yes, is also used by `DefaultMacroContentParser#getCurrentSyntax`). I definitely don't want to create a fake `MacroTransformationContext` just to reuse `getCurrentSyntax` but I guess we could image a `getSyntax(Block)` helper somewhere to avoid having to remember the parameters to pass to `#getFirstBlock`.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/macro/Macro.java
```diff
@@ -70,4 +71,16 @@ public interface Macro<P> extends Comparable<Macro<?>>
      */
     List<Block> execute(P parameters, String content, MacroTransformationContext context)
         throws MacroExecutionException;
+
+    /**
+     * Prepare a {@link MacroBlock} meant to be cached to be executed several times. The goal is to pre-execute
+     * everything that can fit in any context and store it in an annotation of the passed {@link MacroBlock}.
```
michitux, line 77:
```suggestion
     * everything that is independent of any context and store it in an annotation of the passed {@link MacroBlock}.
```

Maybe clearer, not sure.

tmortagne: Yes, it's better.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: I thought `BlockNavigator` was better on that, but when looking at it more closely, it seems to suffer from the same problem with DESCENDANT and DESCENDANT_OR_SELF. Looks like we have some refactoring work to actually make it a proper navigator...

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: After much thought on what was the less bad location and how generic it should be, I just committed a `Block#getSyntaxMetadata`.

tmortagne: WDYT @michitux @surli (and others) ?

michitux: Hm, feels a bit too specific in such a generic location. Maybe it could be a method `<T> T getMetaDataOnSelfOrAncestor(String metaData)`?

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: Created https://jira.xwiki.org/browse/XRENDERING-725.

michitux: Okay, if we have similar recursion elsewhere, the recursion doesn't hurt but I think it would be good to at least make sure that a stack overflow here can't break arbitrary things.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: I tough about this but if we do something too generic then there is not much point compared to directly using `block.getFirstBlock(new MetadataBlockMatcher(MetaData.SYNTAX), Axes.ANCESTOR_OR_SELF);` which you basically reproduced in the name of the method and its parameter.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
tmortagne: Yes, it's probably better, with a comment mentioning https://jira.xwiki.org/browse/XRENDERING-725.

tmortagne: Added some bulletproofing to MacroTransformation#prepare

PR #286, opened 2023-11-08
XRENDERING-736: Macros info, success, warning and error are only distinguished by colors

Sereza7:
## Jira
https://jira.xwiki.org/browse/XRENDERING-736
## PR Changes
* Added the appropriate changes to the block structure of the message macro
## Note
When testing on a default XWiki distribution, those icons use the Silk icon theme despite the selected color theme being Font awesome 4.
## Tests
Passed successfully `mvn clean install -f xwiki-rendering/xwiki-rendering-macros/xwiki-rendering-macro-message/`.

21452-afterPR
Here, we can see that the icons are properly added on a sandbox page with message boxes of the four types.
However, additional style must be provided for them to look better, which will be done in a pull request for `xwiki-platform`.

EDIT: Updated Jira URL

Also see https://github.com/xwiki/xwiki-platform/pull/2590 on `xwiki-platform`


tmortagne:
Would be great to see a screenshot of how it looks like when there is a title.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -73,4 +79,29 @@ public void prepare(MacroBlock macroBlock) throws MacroPreparationException
...
+    public List<Block> execute(BoxMacroParameters parameters, String content, MacroTransformationContext context) 
+        throws MacroExecutionException 
+    {
+        List<Block> boxFundation = super.execute(parameters, content, context);
+        if (boxFundation.size() == 0 || this.iconName == null) {
+            return boxFundation;
+        } else {
+            // Enhance the default box with an icon in the top left.
+            Block defaultBox = boxFundation.get(0);
+            Block iconBlock = new ImageBlock(this.getIconReference(), true);
```
tmortagne, line 99: An actual image inserted in the content does not feel like the cleanest. I would have expected the exact icon to be a CSS (so on platform side) decision based on the class (so, like the color).

Sereza7: We want the icon to be accessible from a screen reader, so unfortunately we can't just add it in the `:before` pseudo-element of the message box.

I thought the best practice would be to add an image in it, but I think I could also just add an empty `Format Block` just like for the title, and somehow fill it with CSS.

However AFAIK, the various iconthemes also have various ways to represent icons in HTML, some use images, some use fonts, ... so I'm not sure CSS only would solve this completely.

Sereza7:
> Would be great to see a screenshot of how it looks like when there is a title.

21452-afterPR2

The title is not rendered when inline. The style displayed here is not final, I plan on making adjustements in some CSS files on xwiki-platform.

Sereza7:
Setting this PR as a draft to add a block around the title, so that it's easier to style.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
michitux: We have unfortunately no way to insert icons by CSS unless I missed something. Regarding the screen reader aspect, that's easy to solve I think by adding a `sr-only` text. But then again, I'm not really sure that this is something we should do in rendering as `sr-only` is not a standard but comes from Bootstrap iirc. Imho it would be cleanest to override these macros in `xwiki-platform` to add proper icon HTML from the icon theme.

Sereza7:
Updated this PR to add a container for the box title: this is a span, with the class 'box-title'.

Here is a display of what this looks like in a XWiki HTML structure:
21452-boxTitleUpdateRendering

xwiki-rendering-macros/xwiki-rendering-macro-box/src/test/resources/macrobox8.test
```diff
@@ -10,9 +10,9 @@
 beginDocument
 beginMacroMarkerStandalone [custombox] [] [something]
 beginGroup [[class]=[box]]
-beginMetaData [[non-generated-content]=[java.util.List<org.xwiki.rendering.block.Block>][parameter-name]=[title]]
```
surli, line 13: Does it mean that the title is not inplace editable anymore? AFAIR those metadata blocks were there on purpose to allow direct edition.

Sereza7: Indeed, without this block, it's not possible to select the title for direct inplace editing. We can only change it through the message box properties.

Sereza7: :+1: Addressed in https://github.com/xwiki/xwiki-rendering/pull/286/commits/dcd57c288fa3f4f0c25f733fd614bde88ee0e82a

surli (approved):

xwiki-rendering-macros/xwiki-rendering-macro-box/src/main/java/org/xwiki/rendering/macro/box/AbstractBoxMacro.java
```diff
@@ -284,12 +287,14 @@ public Block build() throws MacroExecutionException
...
                     // Don't execute transformations explicitly. They'll be executed on the generated content later on.
-                    List<? extends Block> titleBlock = AbstractBoxMacro.this.contentParser.parse(
+                    List<Block> titleContentBlock = AbstractBoxMacro.this.contentParser.parse(
                         titleParameter, context, false, true).getChildren();
-
-                    // put the right metadata around it
-                    ret.addChildren(Collections.singletonList(new MetaDataBlock(titleBlock,
-                        AbstractBoxMacro.this.getNonGeneratedContentMetaData("title"))));
+                    // Put metadata around it so that it's inplace editable
+                    List<Block> titleMetadata = Collections.singletonList(new MetaDataBlock(titleContentBlock,
```
manuelleduc, line 293: Small improvement, slightly shorter and based on java APIs:
```suggestion
                    List<Block> titleMetadata = List.of(new MetaDataBlock(titleContentBlock,
```

Sereza7: :+1: Thanks! 
This function did look odd, but I just reused what was already working to avoid breaking something unexpectedly :)

xwiki-rendering-macros/xwiki-rendering-macro-box/src/test/resources/macrobox8.test
michitux: Shouldn't the title be a block element like `<div>`? I remember that we had a problem that inline elements weren't editable in CKEditor.

Sereza7: AFAICS this works properly with [the changes proposed on xwiki-platform](https://github.com/xwiki/xwiki-platform/pull/2590):

[Demo video](https://up1.xwikisas.com/#tAgXpwwH8OU17qh3nkjTTQ)

Sereza7:
The last changes add a new API to provide icons for rendering. Previously it only supported images. Now this can be extended. This has the aadvantage of allowing support of icon themes that are not based on images, like Font Awesome :)

__
I took this opportunity to update the transformation feature, and not just the AbstractMessage, as they are both very similar.

Here is what a page rendered in XWiki 16.0-SNAPSHOT looks like with the iconTheme being Font awesome:
21452-replacingIcons-FA

Sereza7:
Added some wrapping around the title + content of the block when necessary
Most of the changes are on https://github.com/xwiki/xwiki-platform/pull/2590
21452-updatedStyle-FA

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -19,18 +19,22 @@
...
 import org.xwiki.rendering.macro.MacroPreparationException;
 import org.xwiki.rendering.macro.box.AbstractBoxMacro;
 import org.xwiki.rendering.macro.box.BoxMacroParameters;
 import org.xwiki.rendering.macro.descriptor.DefaultContentDescriptor;
 import org.xwiki.rendering.transformation.MacroTransformationContext;
+import org.xwiki.rendering.transformation.icon.IconProvider;
+
+import javax.inject.Inject;
+import java.util.Collections;
+import java.util.List;
```
michitux, line 37: Wrong import order (see also below).

michitux (changes requested):
Thank you for updating this! I like the new approach with the icon provider. Just some details that need to be fixed. Also, please also open a rendering issue for this, or we should move the issue to rendering. It is fine to have a commit with a XRENDERING-Jira-issue in platform but not the other way around from my understanding.

The update of the icon transformation is nice, but I think it should be its own Jira issue and its own pull request or at least clearly separate commits as while I would backport the box macro to 15.10.x, I wouldn't do that for the icon transformation. The reason for this is that I'm worried that non-images might cause issues with exports. What happens if you export a document with smileys to an office format or LaTeX? I haven't tested this before so I don't know if icons were supported, but in general PNG icons should be much less of an issue than Font Awesome icons. Now I'm not saying that we need to fix this here, just something to think about in general and find a solution in XWiki 16.x.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/286/commits/95bae640b98f40f32cde3abe4216d3e6c7f14a04 :+1:

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/ErrorMessageMacro.java
```diff
@@ -19,12 +19,11 @@
...
 package org.xwiki.rendering.internal.macro.message;
 
-import java.util.Set;
+import org.xwiki.component.annotation.Component;
 
 import javax.inject.Named;
 import javax.inject.Singleton;
-
-import org.xwiki.component.annotation.Component;
+import java.util.Set;
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/286/commits/95bae640b98f40f32cde3abe4216d3e6c7f14a04 :+1:

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/DefaultIconProvider.java
```diff
@@ -0,0 +1,63 @@
...
+public class DefaultIconProvider implements IconProvider
+{
+    private static final List<Class> ICON_CLASS = new ArrayList<Class>(List.of(ImageBlock.class));
+
+    /**
+     * Provides an image icon from its name.
+     * @param iconName the name of the icon needed
+     * @return the block containing the icon
+     */
+    public Block get(String iconName)
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/286/commits/95bae640b98f40f32cde3abe4216d3e6c7f14a04 :+1:

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/transformation/icon/IconProvider.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.List;
+
+/**
+ * Component to provide an icon from its name.
+ *
+ * @version $Id$
+ * @since 15.10.2
+ */
+@Role
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/286/commits/95bae640b98f40f32cde3abe4216d3e6c7f14a04 :+1:

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/IconTransformation.java
```diff
@@ -141,8 +145,8 @@ private Block convertToDeepTree(Block sourceTree, String iconName)
             pointer.addChild(block);
             pointer = block;
         }
-        // Add an image block as the last block
-        pointer.addChild(new ImageBlock(new ResourceReference(iconName, ResourceType.ICON), true));
+        // Add an icon block as the last block
+        pointer.addChild(iconProvider.get(iconName));
```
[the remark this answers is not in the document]
Sereza7: > Are all silk icon names available in all icon themes?
No, but the fallback for any icon theme is Silk, so even if it doesn't find the icon in the mapping, it'll put a Silk icon instead.

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/DefaultIconProvider.java
```diff
@@ -0,0 +1,63 @@
...
+ * Component to provide an icon from its name.
+ *
+ * @version $Id$
+ * @since 15.10.2
+ */
+@Component
+@Singleton
+public class DefaultIconProvider implements IconProvider
+{
+    private static final List<Class> ICON_CLASS = new ArrayList<Class>(List.of(ImageBlock.class));
```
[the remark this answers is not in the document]
Sereza7: Because the iconProvider can send back multiple kinds of icons. See https://github.com/xwiki/xwiki-platform/pull/2590/files#diff-8d2b165ed0bf133c701f7ce2d59119be83117a10967c9257eeb97990143ad7abR73 
This one relies on the default behaviour and can send back RawBlocks and ImageBlocks.

This wouldn't be needed when implementing https://github.com/xwiki/xwiki-rendering/pull/286#discussion_r1428055434

xwiki-rendering-macros/xwiki-rendering-macro-message/pom.xml
```diff
@@ -41,5 +41,10 @@
       <artifactId>xwiki-rendering-macro-box</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+        <groupId>org.xwiki.rendering</groupId>
+        <artifactId>xwiki-rendering-transformation-icon</artifactId>
```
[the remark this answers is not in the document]
Sereza7: Addressed when moving the `IconProvider` to `xwiki-rendering-api` :+1:

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/transformation/icon/IconProvider.java
```diff
@@ -0,0 +1,48 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.rendering.transformation.icon;
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/286/commits/cb8e83dedca82bc78dfb57cddce44302ad5c315d :+1:

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/transformation/icon/IconProvider.java
```diff
@@ -0,0 +1,48 @@
...
+ * @since 15.10.2
+ */
+@Role
+public interface IconProvider
+{
+    /**
+    * Some of the icons might use different iconProviders, so we concatenated all possibilities in a list.
+     * @return The class of the icon block created.
+     */
+    List<Class> getIconClass();
```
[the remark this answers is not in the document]
Sereza7: I introduced the IconBlock in https://github.com/xwiki/xwiki-rendering/pull/286/commits/cb8e83dedca82bc78dfb57cddce44302ad5c315d , removing the need for this API. Therefore I removed all mentions to this method. :+1:

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/IconTransformation.java
```diff
@@ -229,9 +233,9 @@ private void parseTree(List<Block> sourceBlocks)
                     }
                     count++;
                     mappingCursor = mappingCursor.getChildren().get(0);
-                    // If we reach the Image Block then we've found a match!
-                    if (mappingCursor instanceof ImageBlock) {
-                        // Replace the first source block with the image block and remove all other blocks...
+                    // If we reach the Icon Block then we've found a match!
+                    if (iconProvider.getIconClass().contains(mappingCursor.getClass())) {
+                        // Replace the first source block with the icon block and remove all other blocks...
```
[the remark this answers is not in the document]
Sereza7: 1. I introduced the IconBlock class in https://github.com/xwiki/xwiki-rendering/commit/cb8e83dedca82bc78dfb57cddce44302ad5c315d
2.  :+1: See https://github.com/xwiki/xwiki-rendering/pull/286/files#diff-93345a565140d51d15e084f731d9c6c754e61ecae0f8c32403675b07b46acfbcR245

Sereza7:
> Also, please also open a rendering issue for this, or we should move the issue to rendering. It is fine to have a commit with a XRENDERING-Jira-issue in platform but not the other way around from my understanding.

Opened [XRENDERING-736: Macros info, success, warning and error are only distinguished by colors](https://jira.xwiki.org/browse/XRENDERING-736) and updated the title of this PR :)

Note:
TLDR in order to use `xwiki-rendering-api`, one should also build `xwiki-legacy-rendering-api`
 It took me a while to understand how to add my `xwiki-rendering-api` jar to my local distribution. The first step is to build this jar, but then we need to build legacy (making sure it used the proper version of the api...) and create a symlink between the legacy package result and the one in the local distribution.

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/IconTransformation.java
```diff
@@ -229,13 +235,14 @@ private void parseTree(List<Block> sourceBlocks)
...
-                        // Replace the first source block with the image block and remove all other blocks...
+                    // If we reach the Icon Block then we've found a match!
+                    if (mappingCursor instanceof IconBlock) {
+                        // Replace the first source block with the icon block and remove all other blocks...
                         for (int i = 0; i < count - 1; i++) {
                             matchStartBlock.getParent().removeBlock(matchStartBlock.getNextSibling());
                         }
-                        sourceBlock = mappingCursor.clone();
+                        sourceBlock = mappingCursor
+                            .getFirstBlock(AnyBlockMatcher.ANYBLOCKMATCHER, Block.Axes.DESCENDANT).clone();
```
michitux, line 245: Why are you cloning the first child of the icon block? What if the icon block has several children?

michitux (changes requested):

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/IconTransformation.java
Sereza7: The way I defined and used the IconBlock so far is that it contains only one child which is a block containing the whole icon.
---
After thinking about it, there's no reason to exclude other use cases where multiple blocks are given as children, possibly imposing an extra wrapping that's not needed.
---
After looking at the way we parse trees, it's not straigthforward to allow multiple children.
---
I have trouble understanding what this function does, there are a lot of abstract concepts and minimal documentation (and the recursive nature of the function does not make things simpler). With a simple test (1. make sure the default renderer behaves correctly, and make sure a renderer that add a random wordBlock behaves as expected) I couldn't find out an easy change that would return the expected value for tests on both cases...
I think it's okay to assume that all the content in the Icon block is nested inside a singular block.
I'm adding a line in the definition of the IconBlock to highlight this.

Sereza7: We could just merge those children together in a composite block :thinking: looking into it...

xwiki-rendering-api/src/main/java/org/xwiki/rendering/transformation/IconProvider.java
```diff
@@ -0,0 +1,43 @@
...
+@Role
+@Unstable
+public interface IconProvider
+{
+    /**
+     * Provides an image icon from its name.
+     * @param iconName the name of the icon needed
+     * @return the block containing the icon
+     */
+    IconBlock get(String iconName);
```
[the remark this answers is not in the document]
Sereza7: This would open the door to implementing iconProviders that would break the icon transformation feature. I think it's for the best to rely on a more precise class with minimal expectations.

xwiki-rendering-transformations/xwiki-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/IconTransformation.java
Sereza7: Okay, after taking a bit of distance, I figured out I could just wrap all the children block in a compositeBlock if needed. This works properly in both of the situations mentionned above :)

Sereza7: Wrapping the children blocks in a composite block when they are multiple does not work either.

michitux: Why don't you just clone the `IconBlock` itself, i.e., just revert the change on this line?

Sereza7: Somehow it fails at transforming multiple icons one after the other:
https://up1.xwikisas.com/#6lC3HfY35P2Qv2BVGf6LKw

xwiki-rendering-api/src/main/java/org/xwiki/rendering/transformation/IconProvider.java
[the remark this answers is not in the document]
michitux: I have the feeling you kind of misunderstood my original idea. My idea was that the icon transformation wraps whatever block the `IconProvider` returns inside an `IconBlock`, but an `IconBlock` class that is private to the icon transformation. That way, the icon transformation knows also that this `IconBlock` has exactly one child. This `IconBlock` class would not be used anywhere else.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/IconBlock.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+
+/**
+ * This block itself does not have any meaning (it does not have any corresponding rendering stream event) and is just
+ * here to contain any representation of an icon.
+ *
+ * @version $Id$
+ * @since 15.10.4
+ */
+public class IconBlock extends CompositeBlock
```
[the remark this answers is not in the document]
Sereza7: I moved the IconBlock class in the `IconProvider`. This way, it's easily accessible from any class that might need it.
I still don't know if this is a correct place but this looks less wrong to me.

See https://github.com/xwiki/xwiki-rendering/pull/286/commits/8798113c8bcedef52479d5bede79f8138b460c25

From what I understand, it becomes an unstable API, but less open ended, we have a clear context in which to use it.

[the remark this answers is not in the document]
Sereza7: Okay, I removed the `IconBlock` class altogether, replacing it with a static parameter to give to the icon block, in https://github.com/xwiki/xwiki-rendering/pull/286/commits/da1ab07f466768fede0f6c01a5dfaaf0e5b2476b. This has the advantage of not changing the nesting for icon blocks and not breaking the tree traversal algo that was before. 

The disadvantage: 
* using this parameter name-value combination will break the tree traversal and the block will be interpreted like an icon.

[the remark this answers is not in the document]
Sereza7: After seeing your comment at https://github.com/xwiki/xwiki-rendering/pull/286#discussion_r1453718579, I moved this parameter to the iconTransformation class and made it private.
I also updated the name of the parameter to decrease the risk of collision with another feature.
Changes in https://github.com/xwiki/xwiki-rendering/pull/286/commits/5caf98e2287285d41506e940d22ac9ac433d7759

Sereza7:
Changes since https://github.com/xwiki/xwiki-rendering/pull/286#issuecomment-1884942238 :
* Replaced the IconBlock class that was in the API with a block parameter that's defined at the level of icon transformation (since it's only useful in there)
* Moved around the `IconProvider` interface and `DefaultIconProvider` class to the `util` package instead of `transformation` where they were before.
* Misc changes in the different files to fit the changes above.

## Executed tests
As of the end of https://github.com/xwiki/xwiki-rendering/pull/286/commits/af1759b7f264e05f15f6c0ac6a52f1c1317d32a7, I could successfully build:
* `mvn clean install -f xwiki-rendering-api`
* `mvn clean install -f xwiki-rendering-transformations/xwiki-rendering-transformation-icon`
* `mvn clean install -f xwiki-rendering-macros/xwiki-rendering-macro-message`
* `mvn clean install -f xwiki-rendering-macros/xwiki-rendering-macro-box`
which covers all the changes.

Sereza7:
Thanks for your help Michael!
I'll rerun the builds to make sure nothing got broken and keep this topic updated :)

michitux:
As mentioned in the chat, the changes to the icon transformation should be separated from the changes for XRENDERING-736 as they are completely independent, they are just much easier now. The potential problem I see with this change is that export to different formats like office formats or LaTeX won't handle HTML output and thus won't display the icons anymore which could be seen as a regression. I think we need to re-think how we handle icons but regardless how we handle icons in the future, it should be possible to adjust the icon provider to the chosen solution.

Sereza7:
I removed all the changes to the icon transformation in https://github.com/xwiki/xwiki-rendering/pull/286/commits/a451d60904d552efecb6fc6a92e027a4dcdf2045. I created a ticket for this issue: https://jira.xwiki.org/browse/XRENDERING-738 and added the fix discussed here as a comment.

michitux (changes requested):
Thank you, just a few minor suggestions for documentation update.

I fear, following the comment of @vmassol on #2775 there should be a forum proposal that details the UI change with the exact new look that was chosen before this can be merged.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/util/DefaultIconProvider.java
```diff
@@ -0,0 +1,47 @@
...
+import org.xwiki.rendering.util.IconProvider;
+
+import javax.inject.Singleton;
+
+/**
+ * Default component to provide an icon from its name.
+ *
+ * @version $Id$
+ * @since 15.10.6
+ * @since 16.0.0RC1
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/286/commits/6b46afc74e20684431a652bec37cc8f5d6a549c7 :+1:

PR #289, opened 2023-12-12
XRENDERING-667: New lines added when parsing and re-rendering groups in xwiki/2.1

michitux:
* Don't print a new line when closing a group and we're already on a newline.
* Add a test.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-667


tmortagne (approved):

michitux:
This should not be merged as while it fixes the original bug, it introduces another bug which is the removal of a newline at the end of syntax like the following when parsing and rendering it again:

```
(((



Test



)))
```

We're already losing a newline there at the start of the syntax, but with this "fix" we would also lose one at the end. It is not clear to me whose fault this is as we're already loosing newlines in other cases. It is possible that in this case, this paragraph is to blame, and it should print one more newline, see https://jira.xwiki.org/browse/XRENDERING-684 for similar issues with tables and lists.

michitux:
Hopefully better fix in #322.

PR #291, opened 2023-12-14
Update gradle enterprise

ribafish:
- Updated Revved up by Gradle Enterprise badge to say Revved up by Develocity (after GE -> Develocity rename)
- Updated build cache config to only push when authenticated


vmassol:
Thanks!

PR #296, opened 2024-03-27
XRENDERING-736: Macros info, success, warning and error are only distinguished by colors

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XRENDERING-736
# Changes

## Description

* Added the appropriate changes to the block structure of the message macro

## Clarifications

Last attempt at merging a fix for this issue ended up in a lot of CI errors on xwiki-platform. Therefor we reverted changes from https://github.com/xwiki/xwiki-rendering/pull/286 . This PR adds them back. Almost no changes were made on the rendering side since then. Most of the changes were done on the platform side.

# Screenshots & Video

See https://github.com/xwiki/xwiki-rendering/pull/286, no change to the visuals since the end of this PR.

# Executed Tests

Successfully passed `mvn clean install -f xwiki-rendering-macros/xwiki-rendering-macro-message -Pquality,integration-tests,docker;mvn clean install -f xwiki-rendering-macros/xwiki-rendering-macro-box -Pquality,integration-tests,docker; mvn clean install -f xwiki-rendering-api -Pdocker,integration-tests,quality`

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * No backport.
* This PR was made to be applied with https://github.com/xwiki/xwiki-platform/pull/3023


xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/util/DefaultIconProvider.java
```diff
@@ -0,0 +1,47 @@
...
+import org.xwiki.rendering.util.IconProvider;
+
+import javax.inject.Singleton;
+
+/**
+ * Default component to provide an icon from its name.
+ *
+ * @version $Id$
+ * @since 15.10.6
+ * @since 16.1.0RC1
```
michitux, line 36: Needs to be updated.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/util/IconProvider.java
```diff
@@ -0,0 +1,43 @@
...
+import org.xwiki.component.annotation.Role;
+import org.xwiki.rendering.block.Block;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Component to provide an icon from its name.
+ *
+ * @version $Id$
+ * @since 15.10.6
+ * @since 16.1.0RC1
```
michitux, line 31: Needs to be updated.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/util/DefaultIconProvider.java
Sereza7: :+1: Addressed in https://github.com/xwiki/xwiki-rendering/pull/296/commits/ffce49089b48ecb62ff05ea8c064e6c561b49169

xwiki-rendering-api/src/main/java/org/xwiki/rendering/util/IconProvider.java
Sereza7: :+1: Addressed in https://github.com/xwiki/xwiki-rendering/pull/296/commits/ffce49089b48ecb62ff05ea8c064e6c561b49169

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -100,6 +101,16 @@ public List<Block> execute(BoxMacroParameters parameters, String content, MacroT
             }
             // Enhance the default box with an icon as the first element.
             Block iconBlock = iconProvider.get(iconName);
+            // Provide an accessible name besides this icon
+            // This is the responsibility of the message macro and not the iconProvider which should only provide
+            // icons without any semantics
+            if (iconBlock.getClass() == ImageBlock.class) {
+                iconBlock.setAttribute("alt", iconPrettyName);
+            } else if (!iconPrettyName.equals("")) {
+                Block iconAlternative = new FormatBlock(List.of(new RawBlock(iconPrettyName, Syntax.PLAIN_1_0)),
```
michitux, line 110: That's not how you insert plain text. You need to parse the text using a `plain/1.0` parser and insert the resulting blocks. Raw blocks are only displayed in the respective output syntax, i.e., this would only be displayed when rendering the macro to `plain/1.0` syntax.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -73,4 +93,35 @@ public void prepare(MacroBlock macroBlock) throws MacroPreparationException
...
+                defaultBox.setChildren(List.of(boxTextContent));
+            }
+            // Enhance the default box with an icon as the first element.
+            Block iconBlock = iconProvider.get(iconName);
+            // Provide an accessible name besides this icon
+            // This is the responsibility of the message macro and not the iconProvider which should only provide
+            // icons without any semantics
+            if (iconBlock.getClass() == ImageBlock.class) {
+                iconBlock.setAttribute("alt", iconPrettyName);
+            } else if (!iconPrettyName.equals("")) {
```
tmortagne, line 116: `!iconPrettyName.isEmpty()` or `StringUtils.isNotEmpty(iconPrettyName)`

By the way, we generally prefix fields with `this.`.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/296/commits/554112649ff574ae69f7821fbf60a49e5faae5fe :+1:

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
Sereza7: Addressed both of those in https://github.com/xwiki/xwiki-rendering/pull/296/commits/554112649ff574ae69f7821fbf60a49e5faae5fe :+1:

Sereza7:
Addressed all comments, and updated the version numbers.

xwiki-rendering-api/src/main/java/org/xwiki/rendering/util/IconProvider.java
```diff
@@ -0,0 +1,42 @@
...
+@Role
+@Unstable
+public interface IconProvider
+{
+    /**
+     * Provides an icon from its name.
+     * @param iconName the name of the icon needed
+     * @return the block containing the icon
+     */
+    Block get(String iconName);
```
tmortagne, line 41: This always returns inline content ? If yes, it should be more explicit in the API.

michitux: At the moment, the implementation in xwiki-platform always returns a raw block with HTML but as we allow defining icons with wiki syntax in Icon themes, this could be any block that can be generated with XWiki syntax.

vmassol: Sure, my point was mostly about improving the javadoc to give examples (usually examples are the best to explain an api)

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -73,4 +99,42 @@ public void prepare(MacroBlock macroBlock) throws MacroPreparationException
...
+                Block boxTextContent = new GroupBlock(defaultBox.getChildren());
+                defaultBox.setChildren(List.of(boxTextContent));
+            }
+            // Enhance the default box with an icon as the first element.
+            Block iconBlock = this.iconProvider.get(this.iconName);
+            // Provide an accessible name besides this icon
+            // This is the responsibility of the message macro and not the iconProvider which should only provide
+            // icons without any semantics
+            if (iconBlock.getClass() == ImageBlock.class) {
+                iconBlock.setAttribute("alt", this.iconPrettyName);
```
tmortagne, line 121: This feels like something that should eventually be translated, but maybe you have another issue about that ?

tmortagne: OK, I saw you overwritten all the macro on platform side. Not very consistent with the way icon were handled (since you overwrite everything, you could have done the same for icons as you did for the pretty name, not that I find it very clean of course...).

xwiki-rendering-api/src/main/java/org/xwiki/rendering/util/IconProvider.java
Sereza7: > This always returns inline content ? If yes, it should be more explicit in the API documentation.

It would make sense to have this limitation, but so far we didn't need it to be specifically inline so I don't think we need to set a comment about it in the API. If at some point we need a function that returns only inline content, we can always extend the API and add `getInline`.

> I think you should provide some examples in the javadoc about types of Block that can be used: ImageBlock (icons represented by images) + the Block type used for FA.

:+1: done in https://github.com/xwiki/xwiki-rendering/pull/296/commits/c03fabe496852a0f1d5cea38b8c52d83ca3e539d

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
Sereza7: > OK, I saw you overwritten all the macros on platform side. Not very consistent with the way icons were handled (since you overwrite everything, you could have done the same for icons as you did for the pretty name, not that I find it very clean of course...).

Yeah, from what I understood this is related to the discussion we had at https://github.com/xwiki/xwiki-platform/pull/3023#issuecomment-2155077952 . It's not the best, but for now it works and it doesn't lock us into a bad implementation.

tmortagne: The problem with overwriting those macros is that eventually someone will add more stuff to those overwritten macros ("temporary" stuff tend to always stay longer than you initially, though...), and it will get harder and harder to get rid of them. An internal provider would really not cost much, and it does not lock us in any bad implementation either (but is much closer to the logic we eventually want).

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -73,4 +100,43 @@ public void prepare(MacroBlock macroBlock) throws MacroPreparationException
...
+            if (defaultBox.getChildren().size() > 1) {
+                Block boxTextContent = new GroupBlock(defaultBox.getChildren());
+                defaultBox.setChildren(List.of(boxTextContent));
+            }
+            // Enhance the default box with an icon as the first element.
+            Block iconBlock = this.iconProvider.get(this.iconName);
+            // Provide an accessible name besides this icon
+            // This is the responsibility of the message macro and not the iconProvider which should only provide
+            // icons without any semantics
+            String iconPrettyName = iconPrettyNameProvider.getIconPrettyName(this.iconName);
```
tmortagne, line 121: I'm wondering if it would not be much simpler and more logical anyway to just pass the macro id here instead of having to go through a "iconName" variable ?

Also, it would probably make more sense to put the `ApplicationResources.properties` filed with the translations on rendering side since that's where are located the macros those translations are about.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java
```diff
@@ -0,0 +1,39 @@
...
+package org.xwiki.rendering.internal.macro.message;
+
+import org.xwiki.component.annotation.Component;
+
+import javax.inject.Singleton;
+import java.util.Map;
+
+@Component
+@Singleton
+class MacroIconPrettyNameProvider {
```
tmortagne, line 29: I'm surprised checkstyle is OK with this formatting (I would have also expected it to fail because of the missing class javadoc).

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java
```diff
@@ -0,0 +1,39 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.rendering.internal.macro.message;
+
+import org.xwiki.component.annotation.Component;
+
+import javax.inject.Singleton;
+import java.util.Map;
+
+@Component
```
tmortagne, line 27: Missing javadoc.

Sereza7: Sorry, I did forgot to set the PR as a draft. Indeed, it did not pass the build successfully. It was failing on integration-tests and I never even got through quality checks.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java
```diff
@@ -19,21 +19,18 @@
...
-  }
+/**
+ * Provide an alternative text for an icon.
+ *
+ * @version $Id$
+ * @since 16.7.0RC1
+ */
+@Role
+@Unstable
+interface MacroIconPrettyNameProvider
```
tmortagne, line 33: Honestly, you don't need to go through an interface for an internal component. Your previous one was fine, was just missing a role (you can set itself as role in the `@Component`).

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
Sereza7: > I'm wondering if it would not be much simpler and more logical anyway to just pass the macro id here instead of having to go through a "iconName" variable ?

I'm not sure what we could use as a macro id. The macro class would work since we want the same alternative for all the macros of the same class. Since this is internal, we can make the assumption : 1 macro type <-> 1 icon name

> Also, it would probably make more sense to put the ApplicationResources.properties file with the translations on rendering side since that's where are located the macros those translations are about.

Done in https://github.com/xwiki/xwiki-rendering/pull/296/commits/0a423cec12a96c1488e1f1f8a0f7bb7471a8e5c0 :+1: 
I'll keep in mind that we need to create a new weblate project when this PR is merged so that it's easy to contribute translations to this new file.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
Sereza7: Okay, in the latest version, there is no overwriting of the macros at all.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
tmortagne: > I'm not sure what we could use as a macro id.

I don't understand the question, each macro has an id, what is used as role hint.

Sereza7:
Ready for reviews.

Sereza7:
Need to address https://github.com/xwiki/xwiki-rendering/pull/296#discussion_r1707061066

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java
```diff
@@ -0,0 +1,52 @@
...
+        "warning", "Warning"
+    );
+
+    /**
+     * @param macroId the id of the macro whose icon needs an alternative text
+     * @return the alternative text associated to the provided icon
+     */
+    String getIconPrettyName(String macroId) 
+    {
+        return PRETTYNAME_MAPPING.get(macroId);
```
tmortagne, line 50: I'm not sure it really worth having such a mapping. What about `StringUrils.capitalize(macroId)` ? Feels good enough for a fallback we never really use in practice.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
Sereza7: Added in https://github.com/xwiki/xwiki-rendering/pull/296/commits/055972f22dff4437e94aeefeafce54953cce957f that icon provided via this method are always inline. :+1:

xwiki-rendering-api/src/main/java/org/xwiki/rendering/util/IconProvider.java
Sereza7: Added in  https://github.com/xwiki/xwiki-rendering/pull/296/commits/055972f22dff4437e94aeefeafce54953cce957f that this method always returns an inline icon.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java
Sereza7: I went with this solution first, because it's usually bad for accessibility to use programmatic names as descriptions. Here it does not really make a difference so might as well use the macro Id directly :+1: 
Addressed in https://github.com/xwiki/xwiki-rendering/pull/296/commits/75253ad3a2ea2d5e670085a8b024c0a93f35e392

Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/296/commits/75253ad3a2ea2d5e670085a8b024c0a93f35e392 :+1:

Sereza7:
Here is a screenshot of a set of messages I used for manual testing:

image

PR #299, opened 2024-05-02
XRENDERING-705: Footnote macro produces inline content in standalone mode

michitux:
# Jira URL

https://jira.xwiki.org/browse/XRENDERING-705

# Changes

## Description

* "Repair" standalone footnotes by making them inline and wrapping them in a paragraph when processing them in the putFootnotes macro.
* Adapt affected test cases.

## Clarifications

* I decided to put the repair into the putFootnotes macro as it is not possible to change the `inline` property of the macro block otherwise (it cannot be modified and replacing the whole macro block would break the macro transformation).
* I decided to put the paragraph around the macro and not inside the macro (keeping it standalone) as the user experience would be bad - users probably expect to be able to drag the standalone footnote inside some text in the WYSIWYG editor, which won't be possible with a standalone footnote.

# Screenshots & Video

[Screencast_20240502_121725.webm](https://github.com/xwiki/xwiki-rendering/assets/198317/10141432-a7c7-4e5e-9298-3621335acf4f)

# Executed Tests

```
mvn clean install -Pintegration-tests,docker,legacy,standalone,quality
```

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x

### DIFF
--- a/xwiki-rendering-macros/xwiki-rendering-macro-footnotes/src/main/java/org/xwiki/rendering/internal/macro/footnote/PutFootnotesMacro.java
+++ b/xwiki-rendering-macros/xwiki-rendering-macro-footnotes/src/main/java/org/xwiki/rendering/internal/macro/footnote/PutFootnotesMacro.java
@@ -41,6 +41,7 @@ import org.xwiki.rendering.block.LinkBlock;
 import org.xwiki.rendering.block.ListItemBlock;
 import org.xwiki.rendering.block.MacroMarkerBlock;
 import org.xwiki.rendering.block.NumberedListBlock;
+import org.xwiki.rendering.block.ParagraphBlock;
 import org.xwiki.rendering.block.SpaceBlock;
 import org.xwiki.rendering.block.WordBlock;
 import org.xwiki.rendering.block.match.MacroMarkerBlockMatcher;
@@ -192,6 +193,7 @@ public class PutFootnotesMacro extends AbstractMacro<FootnoteMacroParameters>
         // Get the list of footnotes in the document
         macroMarkerBlocks.stream()
             .filter(macro -> FootnoteMacro.MACRO_NAME.equals(macro.getId()))
+            .map(PutFootnotesMacro::repairStandaloneFootnote)
             .map(Footnote::new)
             .forEach(footnote -> footnotes.put(
                 Objects.requireNonNullElseGet(footnote.id, () -> String.valueOf(temporaryId.getAndIncrement())),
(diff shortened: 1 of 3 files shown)


xwiki-rendering-macros/xwiki-rendering-macro-footnotes/src/main/java/org/xwiki/rendering/internal/macro/footnote/PutFootnotesMacro.java
```diff
@@ -228,6 +230,25 @@ public List<Block> execute(FootnoteMacroParameters parameters, String content, M
...
+     * @return the repaired macro marker block
+     */
+    private static MacroMarkerBlock repairStandaloneFootnote(MacroMarkerBlock macroMarkerBlock)
+    {
+        if (macroMarkerBlock.isInline()) {
+            return macroMarkerBlock;
+        } else {
+            // Wrap the macro marker block in a paragraph and make it inline.
+            MacroMarkerBlock result = new MacroMarkerBlock(macroMarkerBlock.getId(), macroMarkerBlock.getParameters(),
+                macroMarkerBlock.getContent(), macroMarkerBlock.getChildren(), true);
```
michitux, line 246: @tmortagne I'm wondering if I should move the attributes of the block, too, any opinions on that?

tmortagne: We don't have any attribute on MacroMarkerBlock right now AFAIK, but I feel it would be more future-proof to keep them on the new MacroMarkerBlock. We might have the same problem in other places where we do this kind of bloc refactoring.

That being said, shouldn't the content of a standalone footnote be standalone from the start ? It's not clear to me why it needs to be repaired.

michitux: There are two things that make the footnote macro special:

1. The initially produced content doesn't stay in the footnote but is moved to the `putFootnotes` macro by the `putFootnotes` macro and new content is set by the `putFootnotes` macro. So initially, the `footnote` macro should always produce inline content as that's what fits better in `putFootnotes` (it puts it inside a list item).
2. While the new content that is added by the `putFootnotes` macro could be made standalone by wrapping it inside a paragraph, I felt that while technically correct, this won't provide a good user experience in the WYSIWYG editor. It doesn't make any sense for a footnote to be standalone and the WYSIWYG editor provides no way to change the footnote back to inline.

Phrased differently, what I would actually want is a way to indicate that the footnote macro doesn't support standalone mode. As this is not exactly common (I think), I did the quick solution to just wrap the footnote macro in a paragraph.

michitux:
I've discarded this solution in favor of the much simpler solution to wrap the footnote content in a paragraph. As proper solution for the WYSIWYG editor, https://jira.xwiki.org/browse/XRENDERING-784 needs to be implemented.

PR #302, opened 2024-06-06
XRENDERING-647: ]]] in a link parameter breaks link syntax

michitux:
# Jira URL

https://jira.xwiki.org/browse/XRENDERING-647

# Changes

## Description

* prepend every matched character with the escape character
* add a new test and adapt the existing test

## Clarifications

* It seems a bit unfortunate that the API allows specifying whole strings without specifying how they should be escaped. I think the solution to prefix every character should work in all cases.

# Screenshots & Video

# Executed Tests

`mvn clean install -Pintegration-tests,docker,legacy,standalone,quality` in `xwiki-rendering`.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x
  * stable-16.4.x


xwiki-rendering-syntaxes/xwiki-rendering-syntax-xwiki20/src/test/resources/xwiki20/specific/escaping/escaping2.test
```diff
@@ -7,8 +7,8 @@
 .#-----------------------------------------------------
 .expect|xwiki/2.0
 .#-----------------------------------------------------
-[[reference||parameter="v~~al~]]u~"e"]]
-[[image:reference||parameter="v~~al~]]u~"e"]]
+[[reference||parameter="v~~al~]~]u~"e"]]
```
tmortagne, line 10: I would have preferred a more optimized version. But I guess you had this in mind too, and it was too complex for not much gain.

michitux: I found a different fix that should be more "optimized" by matching only the first character of each string and the rest using a look ahead as we only escape the first character. This should correctly find all places that need to be escaped.

tmortagne: Great, thanks for spending time on fixing this!

tmortagne (approved):

PR #308, opened 2024-06-14
XRENDERING-749: Deprecated MacroDescriptor's getDefaultCategory() is ignored

tmortagne:
# Jira URL

https://jira.xwiki.org/browse/XRENDERING-749

# Changes

## Description

Fix a retro compatibility breakage introduced by XRENDERING-664.

# Executed Tests

Unit test added in DeprecatedMacroDescriptorTest.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches: all supported branches


pom.xml
```diff
@@ -275,7 +275,19 @@
                    </differences>
                  </revapi.differences>
             -->
-            
+            <revapi.differences>
+              <justification>AspectJ false positives</justification>
```
vmassol, line 279: I'd explain it as "Not a breakage. We fix the backward compatibility issue by adding back the removed code using an Aspect".

tmortagne: This error has nothing to do with the retro compatibility we implement. It basically just complain that the aspect is not `privileged` anymore (since the concept does not really exist in the Java version of AspectJ).

vmassol: hmm ok, I'm surprised that revapi would know about aspects, and it's supposed to execute only on bytecode (where "privileged" doesn't exist.

vmassol: idea: We could add it to our list of annotations that should be ignore by revapi

tmortagne: I guess ignoring the whole `org.aspectj.internal.*` might make sense.

PR #319, opened 2024-12-02
XRENDERING-764: Add an option for messages to be set as status

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XRENDERING-764

# Changes

## Description

* Created a class for the parameters specific to message macros (right now, only the new `isActive`
* Used this new parameter in the macro execution to change the result accordingly.
* Updated tests to fit the new method requirements.

## Clarifications

* Note that when this is merged, a new page of documentation will need to be created to contain doc about all messages that is not for generic boxes.
* The CKEditor macro editor uses its own logic to find what to put under the `more` section. We set the new property as `Advanced`, so that it would only be displayed when it is set as True.

# Screenshots & Video

https://github.com/user-attachments/assets/770fbbc5-b43c-4f81-8cec-6a9586e995a1

In this video, we can see multiple uses of the info macro, with the status property set as true or left as false. Everything behaves as expected.

# Executed Tests

Built the rendering module with: ` mvn clean install -f xwiki-rendering-macros/xwiki-rendering-macro-message/ -Pquality`.
Manual tests with a local instance (see video above).
This PR should not change the default behaviour of the message macros (this added an option but actually never used this option). The only change in UI that could break some tests is the addition of a field in the CKEditor macro editor. This is the last field, so I highly doubt this would break any xpath even if there were some poorly defined ones.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None


tmortagne:
You need to add translation keys for all those new parameters (yes, you added only one, but each macro which expose `MessageMacroParameters` will have its own translation key right with the current system). We tend to forget those too often (me included).

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MessageMacroParameters.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @version $Id$
+ * @since 17.0.0RC1
+ */
+public class MessageMacroParameters extends BoxMacroParameters 
+{
+    /**
+     * @see #isStatus()
+     */
+    private boolean isStatus;
```
tmortagne, line 39: When the getter is `isStatus`, the field name is generally expected to be `status`.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MessageMacroParameters.java
```diff
@@ -0,0 +1,62 @@
...
+import org.xwiki.properties.annotation.PropertyDisplayType;
+import org.xwiki.rendering.macro.box.BoxMacroParameters;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Parameters for the Message macro.
+ *
+ * @version $Id$
+ * @since 17.0.0RC1
+ */
```
michitux, line 33: I think the whole class should be marked as `@Unstable`. Further, macro parameter classes normally aren't internal in XWiki.

tmortagne: Yes, macro parameters being APIs, it's safer to put them under revapi scrutiny.

michitux (changes requested):
As Thomas said, translation keys are missing. Further, you should add least one or two tests that verify that the new parameter works as expected. Apart from this and the other small comments, this looks good, thank you for taking care of this!

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MessageMacroParameters.java
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/319/commits/3f1d240689b391f0db095b2172ff7309464c2262 :+1:

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MessageMacroParameters.java
Sereza7: Addressed in https://github.com/xwiki/xwiki-rendering/pull/319/commits/3f1d240689b391f0db095b2172ff7309464c2262 :+1:

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MessageMacroParameters.java
```diff
@@ -0,0 +1,62 @@
...
+public class MessageMacroParameters extends BoxMacroParameters 
+{
+    /**
+     * @see #isStatus()
+     */
+    private boolean isStatus;
+
+    /**
+     * @since 17.0.0RC1
+     * @return whether or not the current message is a status.
```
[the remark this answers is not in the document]
Sereza7: Added some in https://github.com/xwiki/xwiki-rendering/pull/319/commits/3f1d240689b391f0db095b2172ff7309464c2262 :+1: 

Since it's javadoc and we're not too concerned about brevity, I went and added a lot of info.
Let me know if it's not clear enough.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MessageMacroParameters.java
```diff
@@ -0,0 +1,62 @@
...
+    public boolean isStatus()
+    {
+        return this.isStatus;
+    }
+
+    /**
+     * @since 17.0.0RC1
+     * @param isStatus refers to {@link #isStatus()}
+     */
+    @PropertyDescription("Whether or not this message should be announced as a status.")
```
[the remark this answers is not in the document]
Sereza7: > This needs a much better description as otherwise users will be completely lost when this flag should be set. The description should clearly explain what consequences this has and when it should be used. 

With translations, when do users actually see this description? I improved it a bit to make it more clear but I wanted to avoid it being longer than a full line. 

> Also, I'm wondering if it could make sense to more generically expose a "role" parameter - or is it clear that there is no other role that could be interesting?

There's no other role for such messages that would provide significant added value. There's `alert`, but from what I could see, in pretty much all the use cases it's just as good to use `status` instead :)

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role

Note that for alerts we should have XNotifications.
I just added this status role because in practice I could see a lot of apps and UIs use those message boxes to convey the status of a process. E.g. error box when a query failed,...

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/AbstractMessageMacro.java
```diff
@@ -118,12 +116,15 @@ public void prepare(MacroBlock macroBlock) throws MacroPreparationException
...
     @Override
-    public List<Block> execute(BoxMacroParameters parameters, String content, MacroTransformationContext context) 
+    public List<Block> execute(MessageMacroParameters parameters, String content, MacroTransformationContext context) 
         throws MacroExecutionException 
     {
         List<Block> boxFoundation = super.execute(parameters, content, context);
         if (!boxFoundation.isEmpty() && getIconName() != null) {
             Block defaultBox = boxFoundation.get(0);
+            if (parameters.isStatus()) {
+                defaultBox.setParameter("role", "status");
```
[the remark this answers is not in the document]
Sereza7: This is in the AbstractMessageMacro, which expects an icon name to do anything anyways. The check on the iconName a couple lines ago is just a safegard and should never be failed. When it's failed, the message will just be rendered like a standard box, for which the status parameter does not really make sense.

I don't remember why I added this safegard in `XRENDERING-736: Macros info, success, warning and error are only distinguished by colors` https://github.com/xwiki/xwiki-rendering/commit/3a0a1deed837fe94f2815a9f9c81b404971c1b08 but it doesn't hurt IMO.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/macro/message/MessageMacroParameters.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return whether the current message is an ARIA status.
+     */
+    @Unstable
+    public boolean isStatus()
+    {
+        return this.status;
+    }
+
+    /**
+     * @since 17.0.0RC1
```
michitux, line 56: Wrong version.
```suggestion
     * @since 17.1.0RC1
```

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/macro/message/MessageMacroParameters.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param status refers to {@link #isStatus()}
+     */
+    @PropertyDescription("Whether this message should be set as an ARIA status.")
```
michitux, line 59: I would have expected this to be the same as the translation. But I guess in the end it doesn't matter.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/resources/ApplicationResources.properties
```diff
@@ -23,3 +23,11 @@ rendering.macro.message.icon.alternative.info=Information
 rendering.macro.message.icon.alternative.success=Success
 rendering.macro.message.icon.alternative.warning=Warning
 rendering.macro.message.icon.alternative.error=Error
+rendering.macro.info.parameter.status.name=Status
+rendering.macro.info.parameter.status.description=Use this if your message is generated dynamically or contains information about the current status of a process. This will make sure even assistive technology users are notified of the content of your message.
```
michitux, line 27: I don't really understand the "if your message is generated dynamically" use case. Could you maybe provide an example what you had in mind here? Or do you mean "updated" instead of "generated"?

michitux (approved):
Looks good overall, just a few minor comments.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/macro/message/MessageMacroParameters.java
Sereza7: I talked a bit about this on https://github.com/xwiki/xwiki-rendering/pull/319#discussion_r1946347336 :+1: 

I guess we can still update it later if needed, there's all the info necessary for devs in the comments.

xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/resources/ApplicationResources.properties
Sereza7: OFC it's also important for messages where the content is updated, but usually it would be part of `contains information about the current status of a process`.

Sereza7:
Minor comments addressed :+1:

tmortagne (approved):

PR #330, opened 2025-03-31
XRENDERING-778: Rendering a page with thousands of macros is slow

michitux:
# Jira URL

https://jira.xwiki.org/browse/XRENDERING-778

# Changes

## Description

* Introduce the concept of isolated macros.
* In the macro transformation, use a priority queue to efficiently process macros in order.
* Don't re-scan the XDOM after executing isolated macros.

## Clarifications

This introduces a slight overhead if no macros are isolated. For 5k macros without skin, the overhead seems around 12.16%, so it adds about 14% of execution time in this case. The overhead only depends on the number of macros being in the XDOM not yet executed at the same time, so I think having 5k macros at the top level without any other content is kind of the worst case in terms of overhead.

# Screenshots & Video

No UI changes.

# Executed Tests

```
LANG=C.UTF-8 mvn clean install -Pdocker,legacy,integration-tests,snapshotModules,quality,distribution,flavor-integration-tests,standalone
```

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None for now, seems a bit dangerous.

### DIFF
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
@@ -295,6 +295,9 @@ public class MacroTransformation extends AbstractTransformation implements Initi
     @Inject
     private ErrorBlockGenerator errorBlockGenerator;
 
+    @Inject
+    private IsolatedExecutionConfiguration isolatedExecutionConfiguration;
+
     /**
      * Used to generate Macro error blocks when a Macro fails to execute.
      */
@@ -388,7 +391,9 @@ public class MacroTransformation extends AbstractTransformation implements Initi
                     continue;
                 }
 
-                if (!((Macro<Object>) macro).isExecutionIsolated(macroParameters, macroBlock.getContent())) {
+                if (!((Macro<Object>) macro).isExecutionIsolated(macroParameters, macroBlock.getContent())
+                    && !this.isolatedExecutionConfiguration.isExecutionIsolated(macroBlock.getId()))
+                {
                     priorityMacroBlockMatcher.reset();
                 }
                 newBlocks = ((Macro) macro).execute(macroParameters, macroBlock.getContent(), macroContext);
--- a/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
+++ b/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
@@ -19,10 +19,12 @@
  */
 package org.xwiki.rendering.internal.transformation.macro;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.PriorityQueue;
 
 import javax.inject.Inject;
 import javax.inject.Named;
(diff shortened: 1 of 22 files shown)


manuelleduc:
Do you see the need to add a mechanism to mark a macro as `isExecutionIsolated=true` through external configuration?
This could help improve the performance of macro provided by extensions without having to recompile them, and to move the minimal version to 17.3.0+

michitux:
> Do you see the need to add a mechanism to mark a macro as `isExecutionIsolated=true` through external configuration?
> This could help improve the performance of macro provided by extensions without having to recompile them, and to move the minimal version to 17.3.0+

Sounds like a very interesting suggestion to be able to improve the performance on existing installations without having to modify all macros. Should we have something like `rendering.transformation.macro.additionalIsolatedMacros` (supporting several macro IDs as values) or `rendering.macro.${id}.executionIsolated = true`?

tmortagne:
> especially if we have the same rendering.macro.${id} prefix used for other configurations

We do use this prefix a lot, actually, but for translations right now and not configuration. Still, I think it's a good idea to go with this prefix for configuration too. I can totally image allowing to eventually overwritting most of the macro descriptor (and maybe other macro properties, like the priority) through that kind of properties (which would mean deprecating `rendering.macroCategories`).

Note that we are kind of using this prefix already but for something specific and not for a generic Macro thing: `rendering.macro.velocity.filter` (which control the default filter to use in a `{{velocity}}` macro, and which is an extremely bad idea to change unless you want to break your wiki).

michitux:
We also have `rendering.macro.code.pygments.style` and `rendering.macro.code.source.attachmentMaximumSize`. 
I've added the discussed configuration option, let me know if this is what you imagined. I've added a cache as I wasn't sure how optimized the configuration source is (though I didn't actually benchmark it).

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
```diff
@@ -388,7 +391,9 @@ public void transform(Block rootBlock, TransformationContext context) throws Tra
                     continue;
                 }
 
-                if (!((Macro<Object>) macro).isExecutionIsolated(macroParameters, macroBlock.getContent())) {
+                if (!((Macro<Object>) macro).isExecutionIsolated(macroParameters, macroBlock.getContent())
+                    && !this.isolatedExecutionConfiguration.isExecutionIsolated(macroBlock.getId()))
```
manuelleduc, line 395: To be exhaustive, we could add a unit test for this condition, the rest looks good. Thanks!

tmortagne:
> I've added a cache as I wasn't sure how optimized the configuration source is (though I didn't actually benchmark it).

I guess in theory the main cost is supposed to be the String -> Boolean conversion. But I'm wondering if going thought the restricted provider is not actually a bit more expensive than the conversion (since it goes through the ComponentManager at every call).

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java
michitux: Actually, after writing this, I found this limitation so bad that I changed it. It's now possible for the configuration to override the value provided by the macro in both ways (so the configuration can force a macro to be considered as not-isolated). The use case would be that a macro wrongly indicates that it is isolated, and the configuration would be a workaround until the macro has been fixed.

michitux:
> I guess in theory the main cost is supposed to be the String -> Boolean conversion. But I'm wondering if going thought the restricted provider is not actually a bit more expensive than the conversion (since it goes through the ComponentManager at every call).

The cost after caching the configuration instance seems to be both to get a context component manager to check if there is a specialized converter for `Boolean.class` and then to call the actual convert method. The cost of this, as low as it might be in isolation, was noticeable both with just 5k ID macros and with a real complex page (20ms in both cases, close to the resolution limit of the profiler I think, but it did report it in both cases). I've therefore left the cache.

xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/test/java/org/xwiki/rendering/internal/transformation/macro/IsolatedExecutionConfigurationTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+@ComponentTest
```
manuelleduc, line 39: Missing Javadoc. Though, Javadoc of test methods is usually pretty bland, I know :)

manuelleduc (approved):

tmortagne (approved):

PR #354, opened 2025-08-28
X wiki syntax block renderer test junit5

manuelleduc:
Waiting for 16.10.11 to be released before merging on all branches.


manuelleduc:
Already migrated in the meantime.

PR #380, opened 2026-07-09
[Misc] Use Collection.isEmpty() instead of size() > 0 / size() == 0

manuelleduc:
Generalize over https://github.com/xwiki/xwiki-platform/pull/5795 using the replace structurally feature of intellij


vmassol (approved):
LGTM

PR #434, opened 2026-09-10
XRENDERING-817: Closing macro syntax in verbatim content can close surrounding macro

michitux:
# Jira URL

https://jira.xwiki.org/browse/XRENDERING-817

# Changes

## Description

* Add a new MACRO_VERBATIM_CONTEXT state in the parser that mirrors VERBATIM_CONTEXT so everything inside nested verbatim blocks is treated as macro content.
* Add a MACRO_VERBATIM_LOOKAHEAD guard token to avoid entering the special state when the verbatim is unclosed.
* Track verbatim depth similar to the existing macro depth to track nesting.

New tests, each with both an xwiki/2.0 and an xwiki/2.1 input:
* macro40: the closing marker inside a verbatim block
* macro41: an unterminated "{{{" still closes the macro
* macro42: the marker inside a nested verbatim block, which is what printVerbatimContent produces for verbatim content containing "{{{...}}}"
* macro43: an escaped "~{{/box}}" does not close the macro
* macro44: an edge case regarding the look-ahead which detects closing verbatim even after the macro ends

## Clarifications

* I don't know how relevant this fix actually is, it naturally appeared when fixing XRENDERING-815, so I asked Claude Code to fix it. And it can be easily triggered in the WYSIWYG editor. Still, it feels like an edge case.
* The edge case shown in macro44 feels like the most dangerous part of this change.
* We could also decide to postpone this change to a xwiki/2.2 syntax.

# Screenshots & Video

No UI changes.

# Executed Tests

Whole xwiki-rendering with quality profile (and legacy and standalone).

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None, feels way too dangerous.

### DIFF
--- a/xwiki-rendering-integration-tests/src/test/resources/simple/macros/macro40.test
+++ b/xwiki-rendering-integration-tests/src/test/resources/simple/macros/macro40.test
@@ -1,0 +1,26 @@
+.#-----------------------------------------------------
+.# A verbatim block inside a box macro contains the macro's own closing marker
+.# ({{/box}}). Unlike a parameter value or a reference (macro38/macro39),
+.# verbatim content is literal, it cannot be escaped.
+.# 
+.# Note: a verbatim that starts with a word ({{{Verbatim ...) is accidentally
+.# protected by the lexer's maximal-munch ({{Verbatim {{/box}} is swallowed 
+.# as a nested-macro start), which is why the leading space matters here.
+.#-----------------------------------------------------
+.input|xwiki/2.0
+.#-----------------------------------------------------
+{{box}}
+{{{ {{/box}} }}}
+{{/box}}
+.#-----------------------------------------------------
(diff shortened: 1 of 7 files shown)


tmortagne:
It's a bit risky, as any parser change, but I see you added protection to not break use cases like `{{code}} {{{ {{/code}}` so the likelihood of this causing problems is far lower than the potential fixes.

tmortagne (approved):

michitux:
> It's a bit risky, as any parser change, but I see you added protection to not break use cases like `{{code}} {{{ {{/code}}` so the likelihood of this causing problems is far lower than the potential fixes.

It was Claude Code that had this idea, but yeah, there is a bit of protection **unless** there is another verbatim syntax on the same page, the case of macro44.

